PR #5, opened 2018-01-31
Fix tests on windows

satamas:


serejke:
Thank you
I've applied equivalent changes.

PR #26, opened 2020-04-16
Add Integration section to the README

hsz:
There are two GitHub Actions available in the market that wrap IntelliJ Plugin Verifier to let automate the verification process of the plugins with GitHub Actions CI.

It's worth to promote them so verifier will be used by the wider group of the developers.


thepieterdc:
Appreciate it! =)

serejke:
@hsz Thank you!

PR #46, opened 2020-09-08
Add full json of a Ktor Feature to be stored as blob in Marketplace

Ololoshechkin:


intellij-plugin-structure/structure-ktor/src/main/kotlin/com/jetbrains/plugin/structure/ktor/KtorFeaturePluginManager.kt
```diff
@@ -104,8 +104,8 @@ class KtorFeaturePluginManager private constructor(private val extractDirectory:
               usage = doc.usage!!,
               options = doc.options!!
             )
-          }
-
+          },
+          fullDescriptorJson = Json.stringify(KtorFeatureDescriptor.serializer(), descriptor)
```
AlexanderPrendota, line 108: Do you need? JsonConfiguration.Stable.copy(isLenient = true, ignoreUnknownKeys = true)

intellij-plugin-structure/tests/build.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+  id "org.jetbrains.kotlin.plugin.serialization"
```
AlexanderPrendota, line 2: we use jackson in test(

intellij-plugin-structure/tests/build.gradle
```diff
@@ -12,6 +16,7 @@ dependencies {
   testImplementation 'junit:junit:4.11'
   testImplementation "com.fasterxml.jackson.module:jackson-module-kotlin:2.11.0"
   testImplementation "com.google.jimfs:jimfs:1.1"
-
+  testImplementation "com.google.jimfs:jimfs:1.1"
```
AlexanderPrendota, line 19: why do u need?

intellij-plugin-structure/structure-ktor/src/main/kotlin/com/jetbrains/plugin/structure/ktor/KtorFeature.kt
```diff
@@ -17,7 +17,8 @@ data class KtorFeature(
   override val pluginVersion: String? = null,
   override val pluginId: String? = null,
   val requiredFeatures: List<String> = emptyList(), // Feature IDs.
-  val documentation: KtorFeatureDocumentation? = null
+  val documentation: KtorFeatureDocumentation? = null,
+  val fullDescriptorJson: String? = null
```
AlexanderPrendota, line 21: fullDescriptorJson should be not null and empty.

AlexanderPrendota (changes requested):

intellij-plugin-structure/structure-ktor/src/main/kotlin/com/jetbrains/plugin/structure/ktor/KtorFeature.kt
```diff
@@ -17,7 +17,8 @@ data class KtorFeature(
   override val pluginVersion: String? = null,
   override val pluginId: String? = null,
   val requiredFeatures: List<String> = emptyList(), // Feature IDs.
-  val documentation: KtorFeatureDocumentation? = null
+  val documentation: KtorFeatureDocumentation? = null,
+  val fullDescriptorJson: String = ""
```
AlexanderPrendota, line 21: missing validation

AlexanderPrendota (changes requested):

intellij-plugin-structure/structure-ktor/src/main/kotlin/com/jetbrains/plugin/structure/ktor/KtorFeature.kt
Ololoshechkin: It is being created in one single place and the field gets set there. See `KtorFeaturePluginManager:110`. So it is just a string representation of the json that is already validated by that time

AlexanderPrendota (approved):

PR #53, opened 2020-10-19
Rewrite ktor version; move compatibility utils in this repo

kesarevs:


intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
```diff
@@ -0,0 +1,66 @@
...
+        MAX_BUILD_VALUE - 1
+      } else {
+        build
+      }
+
+      return branch.toLong() * MAX_COMPONENT_VALUE * MAX_BUILD_VALUE + result.toLong() * MAX_COMPONENT_VALUE
+    }
+
+    private fun isNumberOfNines(p: Int): Boolean {
+      for (i in NUMBERS_OF_NINES) {
```
AlexanderPrendota, line 41: NUMBERS_OF_NINES.any { it == p }

AlexanderPrendota (changes requested):
Did u fix tests?

kesarevs:
> Did u fix tests?

These changes do not affect the plugin descriptor structure, so tests are ok.

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
kesarevs: Fixed, thanks.

AlexanderPrendota (approved):

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+        i = i * 10 + 9
+        numbersOfNines.add(i)
+      }
+
+      val result = IntArray(numbersOfNines.size)
+      i = 0
+      while (i < numbersOfNines.size) {
+        result[i] = numbersOfNines[i]
+        i++
+      }
```
Ololoshechkin, line 56: Does `.toTypedArray()` work here?

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+
+      return branch.toLong() * MAX_COMPONENT_VALUE * MAX_BUILD_VALUE + result.toLong() * MAX_COMPONENT_VALUE
+    }
+
+    private fun isNumberOfNines(p: Int) = NUMBERS_OF_NINES.any { it == p }
+
+    private fun initNumberOfNines(): IntArray {
+      val numbersOfNines = ArrayList<Int>()
+      var i = 99999
```
Ololoshechkin, line 44: Why `99999` is a starting point? Why not `999` or  `9999`?

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+      val result = if (build == Integer.MAX_VALUE || isNumberOfNines(build)) {
+        MAX_BUILD_VALUE - 1
+      } else {
+        build
+      }
+
+      return branch.toLong() * MAX_COMPONENT_VALUE * MAX_BUILD_VALUE + result.toLong() * MAX_COMPONENT_VALUE
+    }
+
+    private fun isNumberOfNines(p: Int) = NUMBERS_OF_NINES.any { it == p }
```
Ololoshechkin, line 40: Maybe p.toString().all { it == '9' } would be easier to read without  `initNumberOfNines`?

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
```diff
@@ -0,0 +1,61 @@
...
+
+
+class CompatibilityUtils {
+  companion object {
+    private const val MAX_BRANCH_VALUE = 1000
+    private const val MAX_COMPONENT_VALUE = 10000
+    private const val MAX_BUILD_VALUE = 100000
+    private val NUMBERS_OF_NINES by lazy { initNumberOfNines() }
+
+    fun versionAsLong(vararg components: Int): Long {
```
Ololoshechkin, line 13: I believe Ktor version is always represented by 3 integer numbers: `X.Y.Z`, hence no  `if (components.size >= 3)` is needed and `vararg components` can be just 3 integer arguments

Ololoshechkin (commented):
Could you please provide a bit more context about the usage of `CompatibilityUtils`

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
kesarevs: No, but `toIntArray()` will work. Thanks, that's the really old piece of code, I just copied it from marketplace repo.

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
kesarevs: Because `MAX_BUILD_VALUE = 10000`. That means that we will use five digits for the second component in `version as long.` We use this approach to represent Idea versions as long. So, we want to reuse it with Ktor. Since we are talking about lower bounds, I think that it's not a problem. Are you agree?

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
kesarevs: This is common method that we're using not only for Ktor, but for Idea plugins too. That's why we allow `components.size != 3`. To validate ktor version components we have `KtorVersion.fromString` method.

kesarevs:
> Could you please provide a bit more context about the usage of `CompatibilityUtils`

Yes, of course. This class was copied from marketplace repo. We are using it to convert IDEA multicomponent builds (e.g. `193.2564.0-Snapshot`) to long. Since Ktor versions are similar to IDEA we decided to reuse this approach. This long values will be used to filter plugins/updates in search by compatibility. So, I think we both will be using it.

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
kesarevs: I see your point. The problem is that we have more than 10000 updates and we call this method really often.

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
Ololoshechkin: Ok, sounds good

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/CompatibilityUtils.kt
Ololoshechkin: Ok, sounds good

Ololoshechkin (approved):

PR #59, opened 2020-12-29
Set require features

AlexanderPrendota:


intellij-plugin-structure/structure-ktor/src/main/kotlin/com/jetbrains/plugin/structure/ktor/KtorFeaturePluginManager.kt
```diff
@@ -108,6 +108,7 @@ class KtorFeaturePluginManager private constructor(private val extractDirectory:
               options = doc.options!!
             )
           },
+          requiredFeatures = this.requiredFeatures,
```
Ololoshechkin, line 111: Can we also add validation that this feature id exists in MP?

AlexanderPrendota: yes

PR #61, opened 2021-04-02
Parse optional flag from product descriptor

LChernigovskaya:


serejke:
@LChernigovskaya oops, sorry for not looking into this PR for so long! Missed the notification somehow

PR #223, opened 2021-08-16
Support Sarif format for output

AlexanderPrendota:


.idea/misc.xml
```diff
@@ -5,6 +5,12 @@
     <groovy codeStyle="LEGACY" />
   </component>
   <component name="ExternalStorageConfigurationManager" enabled="true" />
+  <component name="FrameworkDetectionExcludesConfiguration">
```
AlexanderPrendota, line 8: I will remove it

PR #365, opened 2022-02-08
Add ReSharper and Wave versions and the ability to match them

LChernigovskaya:


AlexanderPrendota (approved):

LChernigovskaya:
Sorry for committing after creating the review. I thought about eap versions; they have a string minor component version, like `2022.1.EAP1`. Decided to use build instead for eaps, but the build, for now, has 4 components, and we need to distinguish different eaps (for verifications), so there is just a list of components.

kesarevs (approved):

chrkv (approved):

PR #375, opened 2022-02-22
Add exclude-external-build-classes-selector option

zonder129:
Add exclude-external-build-classes-selector option to have the ability to verify JetBrains plugins without checking external classes. It is useful for verifying of Kotlin plugin, because checking of JPS classes only create false-positives.

### DIFF
--- a/intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/CmdOpts.kt
+++ b/intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/CmdOpts.kt
@@ -44,6 +44,11 @@ open class CmdOpts(
   var externalClassesPrefixes: Array<String> = arrayOf(),
 
   @set:Argument(
+    "exclude-external-build-classes-selector", alias = "ex-selector", description = "Specify this flag if the Plugin Verifier must exclude selector for classes " +
+    "used for the external build processes such as JPS classes bundled into the Kotlin plugin (`/lib/jps`).")
+  var excludeExternalBuildClassesSelector: Boolean = false,
+
+  @set:Argument(
     "subsystems-to-check",
     alias = "subsystems",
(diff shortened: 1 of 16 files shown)


LChernigovskaya:
It seems, `excludeExternalBuildClassesSelector` is missed in the instance creation of `PluginVerifier` in [plugin-verifier-service](https://github.com/JetBrains/intellij-plugin-verifier/blob/4ea69968fbaed488149bf6d9bf12024614247a4f/plugins-verifier-service/src/main/kotlin/org/jetbrains/plugins/verifier/service/service/verifier/VerifyPluginTask.kt#L65)

zonder129:
@LChernigovskaya fixed the issue, thx! The tests are still red though. Shouldn't `dev` version be replaced somehow during tests?

LChernigovskaya:
@zonder129, I think we just need to `publishToMavenLocal` before running the tests. And it seems it doesn't work for a long time. I'll fix it in a new pull request. So, if the tests work locally and a new functionality doesn't require new tests, your request may be merged.

LChernigovskaya (approved):

AlexanderPrendota:
@zonder129 Hi, i dont see any test here, it's ok? We plan a big refactor of this repository in the future and don't want to break your features

LChernigovskaya:
@zonder129 fixes for github tests are in master, you can try to merge them.

zonder129:
@AlexanderPrendota test for this functionality is a tricky thing - we need to download some `org.jetbrains` plugin and test on it, but AFAIK we can't use any real marketplace URLs in tests. Any suggestion on implementation of such a test?

zonder129:
@AlexanderPrendota not actually, we need some code as well to make sure jps classes verfication not generating any compatibility problems

I'm not sure how to do that without sources, which are pretty heavy, but currently I'm trying to emulate such behaviour with custom lightweight plugin (maybe another mock-plugin)

PR #750, opened 2022-12-20
Add ReSharperCompatibilityUtils

LChernigovskaya:


AlexanderPrendota (approved):

intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/version/Version.kt
```diff
@@ -65,4 +67,42 @@ data class ReSharperVersion(val components: List<Int>, override val productCode:
     else {
       this
     }
+
+  fun getLowerVersion(): ReSharperVersion {
+    if (components.size >= 3 && components[2] != 0) {
+      val minorValue = components[2]
+      require(minorValue > 0) {
```
chrkv, line 74: Could you please elaborate on this check? How this can be possible that any of components is lower than 0? Could you please write a test for this scenario?

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/domain/ReSharperVersionTest.kt
```diff
@@ -67,6 +66,78 @@ class ReSharperVersionTest {
...
+
+  @Test
+  fun getLowerVersionMinorNotZeroTest() {
+    val currentVersion = version("RS-2022.3.1")
+    val lowerVersion = currentVersion.getLowerVersion()
+    Assert.assertEquals(2022, lowerVersion.components[0])
+    Assert.assertEquals(3, lowerVersion.components[1])
+    Assert.assertEquals(0, lowerVersion.components[2])
+    Assert.assertEquals(3, lowerVersion.components.size)
+    Assert.assertEquals("RS-2022.3.0", lowerVersion.asString())
```
chrkv, line 98: Is it actually a correct version? Shouldn't it be `2022.2.9`?

chrkv (approved):

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/domain/ReSharperVersionTest.kt
chrkv: Also how `lowerVersion` would be used? Would it be correct to use `2022.2.9`? I mean it could be that the latest `2022.2.x` version of Resharper is `2022.2.4`, for example

intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/version/Version.kt
LChernigovskaya: Yeah, it seems to be redundant, it's strange for them to be lower than 0, thanks. will fix.

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/domain/ReSharperVersionTest.kt
LChernigovskaya: ReSharper range has `isMinIncluded`, `isMaxIncluded` parameters, so when we get, for example, untilLong and max compatible version for update is not included, we need to take a closer lower resharper version. 
I thought about `2022.2.4` or `2022.2.9`, but it seems there is no difference, both of them cover all possible 2022.2 versions, and in case of `2022.2.9` we shouldn't think which minor version is exactly the last.

>Is it actually a correct version? Shouldn't it be `2022.2.9`?
Why do you think it should be `2022.2.9`. The current version is 2022.3.1, so I suppose that the lower should be 2022.3.0. The original compatibility range, for example, is [2022.2, 2022.3.1), so it's changed to [2022.2, 2022.3.0]

PR #885, opened 2023-04-19
MP-5395: Skip Kotlin default methods from internal usage check

bric3:
Kotlin default methods work differently than JVM default methods, as the compiler is generation the default implementation in an inner class `DefaultImpls`, and the kotlin compiler will generate the method in the implementor, even if the actual code doesn't use it.

This PR addresses that by identifying Kotlin default methods in implementation and skip from analyzing these methods for internal types.

Three verifiers had to be patched

- method arguments
- method local vars
- return type

Reference : [MP-5395](https://youtrack.jetbrains.com/issue/MP-5395/Reports-internal-usage-when-extending-an-interface-with-default-method-having-internal-type)

------------

### MP-5395 description :

Take this interface from IntelliJ, where `getPlaceholderCollector` uses the internal type `CodeVisionPlaceholderCollector`

```kotlin
@ApiStatus.Experimental
interface CodeVisionProvider<T> {
  fun precomputeOnUiThread(editor: Editor): T
  fun getPlaceholderCollector(editor: Editor, psiFile: PsiFile?) : CodeVisionPlaceholderCollector? = null
  @JvmDefault
  fun preparePreview(editor: Editor, file: PsiFile) {  }
  // other functions / vals
}
```

And implement other methods but not `getPlaceholderCollector` :
```kotlin
class ReproducerCodeVisionProvider : CodeVisionProvider<Unit> {
    override val defaultAnchor: CodeVisionAnchorKind = CodeVisionAnchorKind.Top
    override val id: String = "reporducer.setting.id"
    override val name: String = "Dummy"
    override val relativeOrderings: List<CodeVisionRelativeOrdering> = emptyList()

    override fun precomputeOnUiThread(editor: Editor) {
        // no-op
    }

    override fun preparePreview(editor: Editor, file: PsiFile) {
        // something useful
    }
}
```

Now the verifier is failing because it sees in the generated bytecode of `ReproducerCodeVisionProvider` a reference to `CodeVisionPlaceholderCollector`. 

```text
Internal API usages (1): 
    #Internal interface com.intellij.codeInsight.codeVision.CodeVisionPlaceholderCollector reference
        Internal interface com.intellij.codeInsight.codeVision.CodeVisionPlaceholderCollector is referenced in boom.ReproducerCodeVisionProvider.getPlaceholderCollector(Editor, PsiFile) : CodeVisionPlaceholderCollector. This interface is marked with @org.jetbrains.annotations.ApiStatus.Internal annotation or @com.intellij.openapi.util.IntellijInternalApi annotation and indicates that the class is not supposed to be used in client code.
```

And indeed looking at the bytecode we can see how kotlin is handling the default method: 

```text
  // access flags 0x1
  public getPlaceholderCollector(Lcom/intellij/openapi/editor/Editor;Lcom/intellij/psi/PsiFile;)Lcom/intellij/codeInsight/codeVision/CodeVisionPlaceholderCollector;
  @Lorg/jetbrains/annotations/Nullable;() // invisible
```

Which turns out to be calling the default.
```text
    INVOKESTATIC com/intellij/codeInsight/codeVision/CodeVisionProvider$DefaultImpls.getPlaceholderCollector (Lcom/intellij/codeInsight/codeVision/CodeVisionProvider;Lcom/intellij/openapi/editor/Editor;Lcom/intellij/psi/PsiFile;)Lcom/intellij/codeInsight/codeVision/CodeVisionPlaceholderCollector;
```

And unfortunately implementing this interface in Java is not possible either because it's not annotated by `@JvmDefault`.

----

The project is targeting Kotlin 1.6 (IJ 222), and it is compiled using Kotlin compiler 1.8.20.

### DIFF
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
@@ -121,7 +121,7 @@ object KotlinMethods {
       return false
     }
 
-    val actualKotlinOwner = methodInsnNode.owner.substring(0, methodInsnNode.owner.length - "\$DefaultImpls".length)
+    val actualKotlinOwner = methodInsnNode.owner.substringBeforeLast("\$DefaultImpls")
 
     // do we want to walk the whole class hierarchy?
     val isAParent = method.containingClassFile.interfaces.any {
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
@@ -7,10 +7,10 @@ import org.objectweb.asm.tree.AbstractInsnNode
 import org.objectweb.asm.tree.MethodInsnNode
 
 object KotlinMethods {
-  private const val capacity = 10
-  private val cache = object : LinkedHashMap<MethodLocation, Boolean>(10) {
+  private const val CAPACITY = 10
+  private val cache = object : LinkedHashMap<MethodLocation, Boolean>(CAPACITY) {
     override fun removeEldestEntry(eldest: MutableMap.MutableEntry<MethodLocation, Boolean>?): Boolean {
-      return size > capacity
+      return size > CAPACITY
     }
   }
 
@@ -64,6 +64,12 @@ object KotlinMethods {
   private fun Method.isKotlinMethodInvokingDefaultImpls(method: Method): Boolean {
     // filter non kotlin classes
     if (!method.containingClassFile.annotations.any { it.desc == "Lkotlin/Metadata;" }) {
+      return false
+    }
+
+    // Sanity check : if the method does not have bytecode
+    // this heuristic cannot run
+    if (instructions.isEmpty()) {
       return false
     }
 
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
@@ -123,7 +123,18 @@ object KotlinMethods {
 
     val actualKotlinOwner = methodInsnNode.owner.substringBeforeLast("\$DefaultImpls")
 
-    // do we want to walk the whole class hierarchy?
+    // Walking the whole class hierarchy is not necessary because
+    // these methods always delegate to the immediate superinterface. E.g.
+    // ```
+    // interface A {
+    //    fun foo() {}
+    // }
+    //
+    // interface B : A
+    //
+    // class C : B
+    // ```
+    // `C.foo` will have a call to `B$DefaultImpls.foo`.
     val isAParent = method.containingClassFile.interfaces.any {
       it == actualKotlinOwner
     }
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
@@ -1,0 +1,69 @@
+package com.jetbrains.pluginverifier.verifiers.method
+
+import com.jetbrains.pluginverifier.verifiers.resolution.Method
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.tree.MethodInsnNode
+
+object KotlinMethods {
+
+
+  /**
+   * Identify a Kotlin default method.
+   *
+   * A kotlin default method is a method that has a default implementation, but kotlin compiles
+   * in the inhertor an override that calls the default implementation (inner) class known as
+   * `DefaultImpls`.
+   *
+   * There's only 3 interesting bytecode, the rest are labels and line numbers.
+   *
+   * ```
+   * // access flags 0x1
+   * public getPlaceholderCollector()Linternal/defaultMethod/AnInternalType;
+   *     @Lorg/jetbrains/annotations/Nullable;() // invisible
+   *     L0
+   *        LINENUMBER 5 L0
+   *        ALOAD 0
+   *        INVOKESTATIC internal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI$DefaultImpls.getPlaceholderCollector (Linternal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI;)Linternal/defaultMethod/AnInternalType;
+   *        ARETURN
+   *     L1
+   *        LOCALVARIABLE this Linternal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI; L0 L1 0
+   *        MAXSTACK = 1
+   *        MAXLOCALS = 1
+   * ```
+   */
+  fun Method.isKotlinDefaultImpl(): Boolean {
+    val method: Method = this
+    // filter non kotlin classes
+    if (!method.containingClassFile.annotations.any { it.desc == "Lkotlin/Metadata;" }) {
+      return false
+    }
+
+    // check the only 3 opcodes in this method are
+    val realOpcodes = method.instructions.filter { it.opcode != -1 } // label or line numbers don't have opcode
+    if (realOpcodes.size != 3
+      || realOpcodes[0].opcode != Opcodes.ALOAD
+      || realOpcodes[1].opcode != Opcodes.INVOKESTATIC
+      || realOpcodes[2].opcode != Opcodes.ARETURN
+    ) {
+      return false
+    }
+
+    val methodInsnNode = realOpcodes[1] as MethodInsnNode
+    if (methodInsnNode.name != method.name || !methodInsnNode.owner.endsWith("\$DefaultImpls")) {
+      return false
+    }
+
+    val actualKotlinOwner = methodInsnNode.owner.substring(0, methodInsnNode.owner.length - "\$DefaultImpls".length)
+
+    // do we want to walk the whole class hierarchy?
+    val isAParent = method.containingClassFile.interfaces.any {
+      it == actualKotlinOwner
+    }
+    if (!isAParent) {
+      return false
+    }
+
+    // The method is a kotlin default method
+    return true
+  }
+}
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
@@ -2,33 +2,47 @@ package com.jetbrains.pluginverifier.verifiers.method
 
 import com.jetbrains.pluginverifier.verifiers.resolution.Method
 import org.objectweb.asm.Opcodes
+import org.objectweb.asm.tree.AbstractInsnNode
 import org.objectweb.asm.tree.MethodInsnNode
 
 object KotlinMethods {
-
 
   /**
    * Identify a Kotlin default method.
    *
    * A kotlin default method is a method that has a default implementation, but kotlin compiles
-   * in the inhertor an override that calls the default implementation (inner) class known as
+   * in the inheritor an override that calls the default implementation (inner) class known as
    * `DefaultImpls`.
    *
-   * There's only 3 interesting bytecode, the rest are labels and line numbers.
+   * Ignoring labels, line numbers, and kotlin intrinsics (to check nullness), there's only a
+   * handful of interesting bytecode that invoke the static method in `{TheInterfaceType}$DefaultImpls`
+   * loading first `this`, then loading parameters if any.
+   *
+   * E.g. with
    *
    * ```
    * // access flags 0x1
-   * public getPlaceholderCollector()Linternal/defaultMethod/AnInternalType;
-   *     @Lorg/jetbrains/annotations/Nullable;() // invisible
-   *     L0
-   *        LINENUMBER 5 L0
-   *        ALOAD 0
-   *        INVOKESTATIC internal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI$DefaultImpls.getPlaceholderCollector (Linternal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI;)Linternal/defaultMethod/AnInternalType;
-   *        ARETURN
-   *     L1
-   *        LOCALVARIABLE this Linternal/defaultMethod/InterfaceWithDefaultMethodUsingInternalAPI; L0 L1 0
-   *        MAXSTACK = 1
-   *        MAXLOCALS = 1
+   * public internalArgsReturningInternal(Linternal/defaultMethod/AnInternalType;Ljava/lang/String;)Linternal/defaultMethod/AnInternalType;
+   *    L0
+   *     ALOAD 1
+   *     LDC "anInternalType"
+   *     INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkNotNullParameter (Ljava/lang/Object;Ljava/lang/String;)V
+   *     ALOAD 2
+   *     LDC "s"
+   *     INVOKESTATIC kotlin/jvm/internal/Intrinsics.checkNotNullParameter (Ljava/lang/Object;Ljava/lang/String;)V
+   *    L1
+   *     LINENUMBER 5 L1
+   *     ALOAD 0
+   *     ALOAD 1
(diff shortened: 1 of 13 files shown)


YannCebron:
test failed

bric3:
For reference running `:intellij-plugin-verifier:verifier-test:build` on `master` and on this branch produces test failures.

<details><summary>Failures of <code>:intellij-plugin-verifier:verifier-test:build</code></summary>

```text
> Task :intellij-plugin-verifier:verifier-test:test FAILED

com.jetbrains.pluginverifier.tests.VerificationTest > classMethod FAILED
    java.lang.IllegalArgumentException at VerificationTest.kt:-1

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin is dynamic if it declars own non-dynamic extension point that is implemented in optional dependency descriptor FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring non dynamic extension point is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring non-dynamic extension in optional dependency config file is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring only predefined extension points can be loaded and unloaded immediately FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > empty plugin can be safely loaded and unload without IDE restart FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring only its own extension points can be loaded and unloaded without restart FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring a group with no ID specified is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring an action with no ID specified is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring actions cant be loaded immediately FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring components is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring only dynamic extension points can be loaded and unloaded without restart FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.DynamicPluginStatusTest > plugin declaring a deep group with no ID specified is not dynamic FAILED
    java.lang.IllegalArgumentException at DynamicPluginStatusTest.kt:319

com.jetbrains.pluginverifier.tests.NoExplicitDependencyOnJavaPluginTest > plugin uses classes of Java plugin but does not declare dependency onto it FAILED
    java.lang.IllegalArgumentException at NoExplicitDependencyOnJavaPluginTest.kt:78

com.jetbrains.pluginverifier.tests.NoExplicitDependencyOnJavaPluginTest > plugin depends on other plugin that refers to Java classes, must not produce warning FAILED
    java.lang.IllegalArgumentException at NoExplicitDependencyOnJavaPluginTest.kt:162

com.jetbrains.pluginverifier.tests.filter.ProblemFilterTest > check only problems matching keep-only-problems filter found FAILED
    java.lang.IllegalArgumentException at ProblemFilterTest.kt:45
```

</details>

bric3:
Yet there's another issue, I'm on it !

bric3:
@YannCebron Fixed the last issue in https://github.com/JetBrains/intellij-plugin-verifier/pull/885/commits/9268f5611f27b988809f37964f4b8eb7960b138c, but unsure whether this one is correct, I'd like guidance.

bric3:
Hi @LChernigovskaya I've updated this branch on the latest master, hopefully it should make it easier to review.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,131 @@
...
+   *     MAXSTACK = 3
+   *     MAXLOCALS = 3
+   * ```
+   *
+   * @see [`JvmDefault` annotation](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/jvm/JvmDefault.kt)
+   */
+  fun Method.isKotlinDefaultMethod(): Boolean {
+    val method: Method = this
+    return cache.computeIfAbsent(method.location) {
+      return@computeIfAbsent isKotlinMethodInvokingDefaultImpls(method)
```
LChernigovskaya, line 60: Sorry, but don't understand why we need `return@computeIfAbsent`

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,131 @@
...
+    // filter non kotlin classes
+    if (!method.containingClassFile.annotations.any { it.desc == "Lkotlin/Metadata;" }) {
+      return false
+    }
+
+    // skip non opcodes, and kotlin intrinsics on arguments
+    val candidateOpcodes = mutableListOf<AbstractInsnNode>()
+    var i = 0
+    do {
+      val currentInsnNode = instructions[i]
```
LChernigovskaya, line 74: Can't `instructions` be an empty list?

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: Otherwise IJ reports a compilation error

> [RETURN_NOT_ALLOWED] 'return' is not allowed here

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: Good point, the code should consider that.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,131 @@
...
+      return false
+    }
+
+    val actualKotlinOwner = methodInsnNode.owner.substring(0, methodInsnNode.owner.length - "\$DefaultImpls".length)
+
+    // do we want to walk the whole class hierarchy?
+    val isAParent = method.containingClassFile.interfaces.any {
+      it == actualKotlinOwner
+    }
+    if (!isAParent) {
```
LChernigovskaya, line 124: `return isAParent`, or it becomes less obvious?

LChernigovskaya: Let's leave it as it is

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/PluginVerifier.kt
```diff
@@ -253,9 +256,13 @@ class PluginVerifier(
       }
     }
 
-    val packageNotFoundProblems = packageToMissingProblems.map { (packageName, missingClasses) ->
-      PackageNotFoundProblem(packageName, missingClasses)
-    }
+    // Kotlin package are supposed to be part of the platform
+    val kotlinPackage = "kotlin\\b(\\..+)?".toRegex()
```
LChernigovskaya, line 260: To be honest, I can't get it — why is it needed and how connected to another part.

bric3: I think I had some issues when runnning the tests locally. And I believe IJ is now shipping kotlin ?
https://plugins.jetbrains.com/docs/intellij/using-kotlin.html#incremental-compilation

I can try remove that change

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
LChernigovskaya: Hm, couldn't get this error, that's why asking :)

bric3: I tried compiling from gralde as well but the compiler still wants the label on the `return`.

```
> Task :intellij-plugin-verifier:verifier-core:compileKotlin FAILED
e: file:///Users/brice.dutheil/opensource/intellij-plugin-verifier/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt:60:7 'return' is not allowed here
```

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,137 @@
...
+    ) {
+      return false
+    }
+
+    val methodInsnNode = candidateOpcodes[candidateOpcodes.lastIndex - 1] as MethodInsnNode
+    if (methodInsnNode.name != method.name || !methodInsnNode.owner.endsWith("\$DefaultImpls")) {
+      return false
+    }
+
+    val actualKotlinOwner = methodInsnNode.owner.substring(0, methodInsnNode.owner.length - "\$DefaultImpls".length)
```
udalov, line 124:
```suggestion
    val actualKotlinOwner = methodInsnNode.owner.substringBeforeLast("\$DefaultImpls")
```

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    // do we want to walk the whole class hierarchy?
```
udalov, line 126: It's not necessary because these methods always delegate to the immediate superinterface. E.g.
```
interface A {
    fun foo() {}
}

interface B : A

class C : B
```
`C.foo` will have a call to `B$DefaultImpls.foo`.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,137 @@
...
+    val isAParent = method.containingClassFile.interfaces.any {
+      it == actualKotlinOwner
+    }
+    if (!isAParent) {
+      return false
```
udalov, line 131: Just want to confirm: is it expected behavior that this check will filter out methods which are generated into the `DefaultImpls` classes of interfaces and delegate to the superinterface?

So for the example above
```
interface A {
    fun foo() {}
}

interface B : A

class C : B
```

For `C.foo`, `isKotlinMethodInvokingDefaultImpls` will return true, which is good.

But there's also `B$DefaultImpls.foo`, for which `isKotlinMethodInvokingDefaultImpls` will return false because `B$DefaultImpls` does not have any superclasses, and this is expected, right?

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -0,0 +1,137 @@
...
+    if (instructions.isEmpty()) {
+      return false
+    }
+
+    // skip non opcodes, and kotlin intrinsics on arguments
+    val candidateOpcodes = mutableListOf<AbstractInsnNode>()
+    var i = 0
+    do {
+      val currentInsnNode = instructions[i]
+      if (currentInsnNode.opcode == -1) {
```
udalov, line 81: I don't know whether this instruction list contains debug info, but if it does, maybe the whole loop can be simplified to skipping until the first LINENUMBER instruction?

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: Thank you I will update the comment.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: > But there's also `B$DefaultImpls.foo`, for which `isKotlinMethodInvokingDefaultImpls` will return false because `B$DefaultImpls` does not have any superclasses, and this is expected, right?

Ah I didn't thought about this, indeed looking at `B$DefaultImpls`' bytecode, super is `Object` and doesn't have interfaces. 

The `B$DefaultImpls.foo` implementation
```
// access flags 0x9
public static foo(Linternal/defaultMethod/B;)Linternal/defaultMethod/AnInternalType;
  // annotable parameter count: 1 (invisible)
  @Lorg/jetbrains/annotations/NotNull;() // invisible, parameter 0
 L0
  LINENUMBER 3 L0
  ALOAD 0
  CHECKCAST internal/defaultMethod/A
  INVOKESTATIC internal/defaultMethod/A$DefaultImpls.foo (Linternal/defaultMethod/A;)Linternal/defaultMethod/AnInternalType;
 L1
  LINENUMBER 10 L1
  ARETURN
 L2
  LOCALVARIABLE $this Linternal/defaultMethod/B; L0 L2 0
  MAXSTACK = 1
  MAXLOCALS = 1
```

I see two cases:
1. If both interfaces `A` and `B` are part of IDEA, then it won't be an issue, because `B` / `B$DefaultImpls` shouldn't be analyzed by the verifier.
2. However if the plugin has an in-between-interface `I` that extends `B`, then this might be an issue, because `I$DefautImpls` will not be filtered out.

Do you think I should introduce additional checks for this specific case, the class is named `xxx$DefaultImpls` and calls another `...$DefaultImpls` ?
I'm ok with that.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: I don't think all plugins have debug info, I know that some are using obfuscators like pro-guard that may remove these (but of course keep parent symbols intact).
I think it's safer to assume debug info is not always present.

bric3:
FYI, just copy/pasting a replay from @chashnikov 



> > The project is targeting Kotlin 1.6 (IJ 222)
> 
> Note, that this problem is actual only for the old implementation of default method in interfaces in Kotlin. If `@JvmDefault` annotation is used, the Kotlin compiler generates default implementation directly in the interface, and won't generate code in the implementing classes. And in 2022.3 builds (starting from [this commit](https://github.com/JetBrains/intellij-community/commit/45d44ea1466ed0d7b5faeab1ad13a7c573fe1408)) we use `-Xjvm-default=all` compiler option to force using JVM default methods for all Kotlin interfaces in IntelliJ sources. So this problem is actual only for 2022.2 and older versions.

And I replied I believe this is still reasonable to address the matter for older versions.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
bric3: @LChernigovskaya Fantastic, I would definitely like to support case 2 as I think it's doable. I just didn't yet started.

bric3: See #976

PR #919, opened 2023-05-05
Edu programming lang field

kesarevs:
https://youtrack.jetbrains.com/issue/MP-5460/Extend-programmingLanguage-field-for-edu-plugins


intellij-plugin-structure/structure-edu/src/main/kotlin/com/jetbrains/plugin/structure/edu/EduPlugin.kt
```diff
@@ -22,7 +22,8 @@ data class EduPlugin(
   override val thirdPartyDependencies: List<ThirdPartyDependency> = emptyList(),
   val descriptorVersion: Int? = null,
   val language: String? = null,
-  val programmingLanguage: String? = null,
```
zbsmirnova, line 25: Shouldn't we keep this field for backwards compatibility with old EduPlugin versions?

kesarevs: I think no. This is an internal object we use after plugin descriptor was parsed. For backward compatibility we still have old field in `EduPluginDescriptor` which is used while plugin descriptor is been parsed. 

I also added a separate test to check that backward compatibility is working correctly

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/edu/mock/EduInvalidPluginsTest.kt
```diff
@@ -58,9 +58,18 @@ class EduInvalidPluginsTest(fileSystemType: FileSystemType) : BasePluginManagerT
...
   fun `programming language is not specified`() {
-    checkInvalidPlugin(PropertyNotSpecified(PROGRAMMING_LANGUAGE)) { programmingLanguage = null }
-    checkInvalidPlugin(PropertyNotSpecified(PROGRAMMING_LANGUAGE)) { programmingLanguage = "" }
-    checkInvalidPlugin(PropertyNotSpecified(PROGRAMMING_LANGUAGE)) { programmingLanguage = "\n" }
+    checkInvalidPlugin(PropertyNotSpecified(PROGRAMMING_LANGUAGE_ID)) {
+      programmingLanguageId = null
+      programmingLanguage = null
+    }
+    checkInvalidPlugin(PropertyNotSpecified(PROGRAMMING_LANGUAGE_ID)) {
+      programmingLanguageId = null
```
chrkv, line 66: I guess `""` should be here instead of `null`

chrkv (approved):

PR #933, opened 2023-05-11
Increase severity for "does not declare explicit dependency on Java plugin" to an error

novotnyr:
- Migrate warning to error
- Move verification logic to a separate class. This decouples the verification context from verification business logic.

https://youtrack.jetbrains.com/issue/MP-5434

### DIFF
--- a/intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/javaPlugin/UndeclaredDependencyOnJavaPluginProblem.kt
+++ b/intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/javaPlugin/UndeclaredDependencyOnJavaPluginProblem.kt
@@ -1,26 +1,22 @@
-/*
- * Copyright 2000-2020 JetBrains s.r.o. and other contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
- */
-
-package com.jetbrains.pluginverifier.warnings
+package com.jetbrains.pluginverifier.usages.javaPlugin
 
 import com.jetbrains.pluginverifier.results.presentation.ClassGenericsSignatureOption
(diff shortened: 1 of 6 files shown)


YannCebron (approved):

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/javaPlugin/UndeclaredDependencyOnJavaPluginProblem.kt
```diff
@@ -1,26 +1,22 @@
...
 
 import com.jetbrains.pluginverifier.results.presentation.ClassGenericsSignatureOption
 import com.jetbrains.pluginverifier.results.presentation.ClassOption
 import com.jetbrains.pluginverifier.results.presentation.formatClassLocation
+import com.jetbrains.pluginverifier.results.problems.CompatibilityProblem
 import com.jetbrains.pluginverifier.usages.formatUsageLocation
-import com.jetbrains.pluginverifier.usages.javaPlugin.JavaPluginClassUsage
+import java.util.*
 
-data class NoExplicitDependencyOnJavaPluginWarning(
```
LChernigovskaya, line 13: Why don't still use `data class` here?

Maybe we can make all problems `data class`es (but in a separate branch of course). What do you think?

novotnyr: Descendants of `CompatibilityProblem` are not declared as data classes anywhere. This is to be consistent.

On the other hand, _compatibility problems_ just hold data, so they are a good candidates. Let's do that on a separate branch.

LChernigovskaya (approved):

PR #993, opened 2023-09-03
Typo in argument name in README.md

t4lz:
The documentation says `-ignore-problems` but [it seems like](https://github.com/JetBrains/intellij-plugin-verifier/blob/639ae7ebbc433a81dd956564be57e4514d504ea7/intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/CmdOpts.kt#L17C1-L17C1) the actual name of that argument is `-ignored-problems`.


novotnyr:
Thank you for spotting this!

PR #1007, opened 2023-09-27
Report an error when the plugin is not v2 and doesn't have dependencies

LChernigovskaya:


LChernigovskaya:
@YannCebron , we already have this check on MP side, and don't allow to upload the plugin without dependencies. We just decided to move this check in structure library, because an IJ team uses this library for checking the plugins before uploading. https://youtrack.jetbrains.com/issue/MP-5948

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
```diff
@@ -589,17 +589,21 @@ internal class PluginCreator private constructor(
 
   private fun validatePlugin(plugin: IdePluginImpl) {
     val dependencies = plugin.dependencies
+    if (!plugin.isV2) {
+      if (dependencies.size == 0) {
```
novotnyr, line 593: isEmpty()?

novotnyr (approved):

PR #1020, opened 2023-10-23
Clarify `-runtime-dir` switch semantics

novotnyr:
The CLI `-runtime-dir` switch is misleading.

Currently, 
```
The path to the directory containing Java runtime jar. If not specified, the JDK from 'JAVA_HOME' will be chosen.
```
However, this is not right and it prevents the non-default JDKs being properly used. In addition, testing plugins against IDEs and separate JDKs is impossible.

## Suggested change

1. Use the `-runtime-dir` as a path to the JDK location (in the same spirit as `JAVA_HOME`). If this is not specified, then:
2. Discover JDK in the embedded IDE. If this is not available, then:
3. Fallback to `JAVA_HOME`.

See [MP-6028](https://youtrack.jetbrains.com/issue/MP-6028)

### DIFF
--- a/intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/OptionsParser.kt
+++ b/intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/OptionsParser.kt
@@ -106,7 +106,11 @@ object OptionsParser {
   }
 
   fun createIdeDescriptor(idePath: Path, opts: CmdOpts): IdeDescriptor {
-    val defaultJdkPath = opts.runtimeDir?.let { Paths.get(it) }
+    val defaultJdkPath = opts.runtimeDir?.let {
+      Paths.get(it).also {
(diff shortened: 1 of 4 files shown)


intellij-plugin-verifier/verifier-cli/src/main/java/com/jetbrains/pluginverifier/options/OptionsParser.kt
```diff
@@ -106,7 +106,11 @@ object OptionsParser {
   }
 
   fun createIdeDescriptor(idePath: Path, opts: CmdOpts): IdeDescriptor {
-    val defaultJdkPath = opts.runtimeDir?.let { Paths.get(it) }
+    val defaultJdkPath = opts.runtimeDir?.let {
+      Paths.get(it).also {
+        LOG.info("Using Java runtime from $it")
```
YannCebron, line 111: I assume previous `print` was a bad choice vs proper logging

YannCebron (approved):

PR #1023, opened 2023-10-26
Collect telemetry data when verifying plugins

novotnyr:
- Establish telemetry framework for plugin descriptor phase and bytecode verification phase
- Gather telemetry on plugin descriptor parsing duration
- Gather telemetry on plugin byte verification duration
- Gather telemetry about number of verified classes in the plugin
- Measure plugin size on file-based plugins
- Make reportage classes report telemetry statistics

### DIFF
--- a/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/telemetry/MutablePluginTelemetry.kt
+++ b/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/telemetry/MutablePluginTelemetry.kt
@@ -8,7 +8,7 @@ import java.time.Duration
 
 const val PLUGIN_ID = "plugin.id"
 const val PLUGIN_VERSION = "plugin.version"
-const val PLUGIN_SIZE = "plugin.size"
+const val ARCHIVE_FILE_SIZE = "plugin.archive.file.size"
 const val PARSING_DURATION = "plugin.parsing.duration"
 const val PLUGIN_VERIFICATION_TIME = "plugin.verification.duration"
 const val PLUGIN_VERIFIED_CLASSES_COUNT = "plugin.verification.verified.classes.count"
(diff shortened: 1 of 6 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
```diff
@@ -892,3 +896,10 @@ internal class PluginCreator private constructor(
       is PluginCreationFail -> this.errorsAndWarnings.filter { it.level === ERROR }
     }
 }
+
+private fun PluginCreationResult<IdePlugin>.add(telemetry: PluginTelemetry): PluginCreationResult<IdePlugin> {
```
YannCebron, line 900: why don't we add telemetry? it could still contain useful (partial) results?

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/utils/FileUtil.kt
```diff
@@ -155,3 +158,6 @@ val Path.extension: String
 
 val Path.length: Long
   get() = Files.size(this)
+
+val Path.pluginSize: Bytes
```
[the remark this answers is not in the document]
novotnyr: Set to `archiveFileSize`.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
novotnyr: Adding a telemetry to the `PluginCreationFail` is quite a large change. Such creation result does not carry plugin metadata (only `File` and the corresponding descriptor errors). Adding this will complicate this PR.

Note that this will carry only the barebones telemetry - plugin archive size, and descriptor parsing duration.

Let's do that in a subsequent PR.

PR #1039, opened 2023-11-27
Use custom name field for edu sections and lessons

kesarevs:


intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/edu/mock/EduPluginMockTest.kt
```diff
@@ -222,6 +223,67 @@ class EduPluginMockTest(fileSystemType: FileSystemType) : BasePluginManagerTest<
...
+      "Section's title should be equal to custom name if it's present",
+      section.customName, stats.sections.firstOrNull()?.title
+    )
+
+    assertEquals("Lesson's title should be equal to custom name if it's present",
+      lesson.customName, stats.lessons.firstOrNull()
+    )
+  }
+
+  @Test
```
LChernigovskaya, line 256: Maybe add a test with a default value for custom name? To check if there is no value pass for `EduItem`, `title` will be used

kesarevs: I did it [here](https://github.com/JetBrains/intellij-plugin-verifier/pull/1039/files#diff-ec8e6629bd4740150875a4cdb1dab75e4adef20c7533eb8a1ced7951a748edf5R257), isn't it?

PR #1044, opened 2023-12-06
Do not search for additional logos when primary logo is not found

novotnyr:
`structure`:
- Do not search for additional logos when primary logo is not found

Tests:
- Add customization of plugin creation in tests, especially when creating via customized factory methods of `IdePluginManager`

Terminology: plugin icon and plugin logo are used interchangeably. The official terminology is [plugin logo](https://plugins.jetbrains.com/docs/intellij/plugin-icon-file.html).


intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/jar/PluginJar.kt
```diff
@@ -21,6 +23,8 @@ val PLUGIN_XML_RESOURCE_PATH = META_INF + File.separator + PLUGIN_XML
 private val THIRD_PARTY_LIBRARIES_FILE_NAME = "dependencies.json"
 
 class PluginJar(private val jarPath: Path, private val jarFileSystemProvider: JarFileSystemProvider = DefaultJarFileSystemProvider()): AutoCloseable {
+  private val LOG: Logger = LoggerFactory.getLogger(PluginJar::class.java)
```
YannCebron, line 26: AFAIU LOG instance could be top-level to avoid instantiation for every instance

YannCebron (changes requested):

intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/jar/PluginJar.kt
novotnyr: Fixed.

By the way, [SLF4J FAQ](https://www.slf4j.org/faq.html#declared_static) says neither approach is better than the other one.

YannCebron: we have this policy for loggers in IJ for a reason

YannCebron (approved):

PR #1046, opened 2023-12-11
Deprecate Plugin Verifier Service

novotnyr:
_Plugin Verifier Service_ has been replaced by a different implementation.

- Mark references in READMEs as deprecated
- Remove Dependabot monitoring


LChernigovskaya:
Small note: `FeatureExtractorService` from `plugins-verifier-service` is still used by Marketplace.

PR #1047, opened 2023-12-11
Enable ignoring of specific plugin problems

novotnyr:
Some plugin problems need to be ignored. An example issue is [MP-5911](https://youtrack.jetbrains.com/issue/MP-5911) _Not expected plugin verifier warning for plugin ID containing `intellij`_.

- Adjust `LevelRemappingPluginCreationResultResolver` to support _ignoring_ of specific plugin problems
- Improve syntax on various modes of level remapping.
- Add unit tests for various combinations of remapped plugin problems.
- Simplify remapping levels of plugin problem collections
- Remove ability to classify a single plugin problem as it is not used anywhere.

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
@@ -9,11 +9,11 @@ import com.jetbrains.plugin.structure.intellij.plugin.IdePlugin
 import kotlin.reflect.KClass
 
 class LevelRemappingPluginCreationResultResolver(private val delegatedResolver: PluginCreationResultResolver,
-                                                 additionalLevelRemapping: Map<KClass<*>, PluginProblem.Level> = emptyMap()
+                                                 additionalLevelRemapping: Map<KClass<*>, RemappedLevel> = emptyMap()
   ) : PluginCreationResultResolver {
 
-  private val remappedLevel: Map<KClass<*>, PluginProblem.Level> = additionalLevelRemapping +
-    mapOf(ForbiddenPluginIdPrefix::class to PluginProblem.Level.WARNING)
+  private val remappedLevel: Map<KClass<*>, RemappedLevel> = additionalLevelRemapping +
+    warning<ForbiddenPluginIdPrefix>()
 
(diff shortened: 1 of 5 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/RemappedPluginProblems.kt
```diff
@@ -0,0 +1,21 @@
+package com.jetbrains.plugin.structure.intellij.problems
+
+import com.jetbrains.plugin.structure.base.problems.PluginProblem
+import kotlin.reflect.KClass
+
+sealed class RemappedLevel
+
+object IgnoredLevel : RemappedLevel()
+class StandardLevel(val originalLevel: PluginProblem.Level) : RemappedLevel()
```
YannCebron, line 9: `data class`?

YannCebron (approved):

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
```diff
@@ -9,11 +9,11 @@ import com.jetbrains.plugin.structure.intellij.plugin.IdePlugin
...
 
 class LevelRemappingPluginCreationResultResolver(private val delegatedResolver: PluginCreationResultResolver,
-                                                 additionalLevelRemapping: Map<KClass<*>, PluginProblem.Level> = emptyMap()
+                                                 additionalLevelRemapping: Map<KClass<*>, RemappedLevel> = emptyMap()
   ) : PluginCreationResultResolver {
 
-  private val remappedLevel: Map<KClass<*>, PluginProblem.Level> = additionalLevelRemapping +
-    mapOf(ForbiddenPluginIdPrefix::class to PluginProblem.Level.WARNING)
+  private val remappedLevel: Map<KClass<*>, RemappedLevel> = additionalLevelRemapping +
+    warning<ForbiddenPluginIdPrefix>()
```
LChernigovskaya, line 16: Why not `ignore` and why not to add `TemplateWordInPluginId` also here as a default remapping? Otherwise, I think we shouldn't have default remapping at all and let the client define remappings.

novotnyr: Removed any of the default remappings. Clients should decide which mappings with corresponding levels they should use.

LChernigovskaya (approved):

PR #1053, opened 2024-01-11
Introduce Gradle Changelog Plugin

novotnyr:
- Introduce [Gradle Changelog Plugin](https://github.com/JetBrains/gradle-changelog-plugin)
- Consolidate version names in Changelog (remove `v` prefix in recent versions)


intellij-plugin-verifier/build.gradle.kts
```diff
@@ -129,3 +130,9 @@ tasks {
   }
 }
 
+changelog {
+  version = projectVersion.also { println(it) }
```
YannCebron, line 134: println?

novotnyr: Leftover, removed as this is not necessary.

PR #1055, opened 2024-01-23
Improve presentation of dynamic status of a plugin in CLI

novotnyr:
- Use separate category for restrictions
- Explicitly emit a list of violated restrictions
- Address both plain text CLI and Markdown output

Sample stdout/CLI output:

```
Deprecated API usages (1): 
    #Deprecated method com.intellij.ide.ApplicationInitializedListener.componentsInitialized() is overridden
        Deprecated method com.intellij.ide.ApplicationInitializedListener.componentsInitialized() : void is overridden in class com.github.novotnyr.nondynamic.MyApplicationInitializedListener

Dynamic Plugin Eligibility (negative due to 2 restrictions):
    Plugin probably cannot be enabled or disabled without IDE restart
    #It declares non-dynamic extensions: `com.intellij.applicationInitializedListener`
    #It declares a group without 'id' specified
```


LChernigovskaya (approved):

YannCebron (approved):
AFAIU no change in text to consider for Plugin DevKit highlighting

PR #1074, opened 2024-03-17
Rework plugin problem level remapping when reclassifying structural plugin descriptor problems

novotnyr:
- Fix case when reclassified plugin problem level might not be used properly, leading to a plugin descriptor discovery issue. 
- Unwrap plugin problem which have reclassified problem level.
- Introduce a special case of _invalid plugin_ which contains seemingly invalid plugin descriptor data. In some scenarios, even plugin with core plugin errors might be treated as valid due to relaxed rules. This is often a case with JetBrains plugins.
Before this improvement, a Plugin creation result resolver was not able to reclassify such basic validation problems.
- Introduce a set of plugin problem remappings for JetBrains plugins.

See [MP-6388](https://youtrack.jetbrains.com/issue/MP-6388)

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
@@ -219,6 +219,24 @@ internal class PluginCreator private constructor(
     plugin.originalFile = originalFile
   }
 
+  /**
+   * Create an instance of an invalid plugin with the most basic information.
+   *
+   * This allows creating a bare-bones plugin with invalid data.
+   * Plugin problems associated with this invalid data might be reclassified by the [problem resolver](#problemResolver).
+   */
+  private fun newInvalidPlugin(bean: PluginBean, document: Document): InvalidPlugin {
+    return InvalidPlugin(document).apply {
+      pluginId = bean.id?.trim()
+      pluginName = bean.name?.trim()
+      bean.vendor?.let {
+        vendor = if (it.name != null) it.name.trim() else null
+        vendorUrl = it.url
+        vendorEmail = it.email
+      }
+    }
+  }
+
   private fun IdePluginImpl.setInfoFromBean(bean: PluginBean, document: Document) {
     pluginName = bean.name?.trim()
     pluginId = bean.id?.trim() ?: pluginName
@@ -656,7 +674,7 @@ internal class PluginCreator private constructor(
     val document = resolveXIncludesOfDocument(originalDocument, documentName, pathResolver, documentPath) ?: return
     val bean = readDocumentIntoXmlBean(document) ?: return
     validatePluginBean(bean, validateDescriptor)
-    if (hasErrors()) {
+    if (hasPluginBeanErrors(bean, document)) {
       return
     }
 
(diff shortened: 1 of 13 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
```diff
@@ -11,7 +11,8 @@ import com.jetbrains.plugin.structure.intellij.plugin.IdePlugin
 import kotlin.reflect.KClass
 
 class LevelRemappingPluginCreationResultResolver(private val delegatedResolver: PluginCreationResultResolver,
-                                                 additionalLevelRemapping: Map<KClass<*>, RemappedLevel> = emptyMap()
+                                                 additionalLevelRemapping: Map<KClass<*>, RemappedLevel> = emptyMap(),
```
LChernigovskaya, line 14: It looks like just `levelRemapping` not `additional` now :)

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
```diff
@@ -734,6 +752,31 @@ internal class PluginCreator private constructor(
...
+   * allowed field values in the plugin descriptor.
+   *
+   * @see validatePluginBean
+   */
+  private fun hasPluginBeanErrors(bean: PluginBean, document: Document): Boolean {
+    if (problems.isEmpty()) {
+      return false
+    }
+    val invalidPlugin = newInvalidPlugin(bean, document)
+    return problemResolver.classify(invalidPlugin, problems).any {
```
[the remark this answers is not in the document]
novotnyr: `PluginCreator` is actually using a two-phase validation.

1. In the first phase, the basic plugin descriptor properties are validated: mandatory fields, field names, plugin ID rules.
2. In the second phase, the complex plugin descriptor properties are checked: dependencies, modules etc.

When the first phase emits any problems, the second phase will not occur.

Originally, the Problem Resolver was not able to do anything in the first phase, as the plugin instance was not even created (there were only raw XML data and a `PluginBean` which is essentially a type-safe mirror of the XML).

Hence this _invalid plugin_. If the first phase fails, at least an invalid plugin is created. Then such plugin can passed down to the problem resolver which might remap even first phase plugin problems into non-problems.

For now, only the minimalistic info is passed down in the invalid plugin (ID, vendor, name), so the JetBrains plugin problem resolver can decide which JetBrains-specific issues will be remapped.

This might sound convoluted, but it can cover complex scenarios with multiple sets of rules.

LChernigovskaya (approved):

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
```diff
@@ -58,7 +59,13 @@ class LevelRemappingPluginCreationResultResolver(private val delegatedResolver:
   }
 
   private fun remapPluginProblemLevel(pluginProblem: PluginProblem): PluginProblem?{
-    return when (val remappedLevel = remappedLevel[pluginProblem::class]) {
+    val problem = if (unwrapRemappedProblems && pluginProblem is ReclassifiedPluginProblem)  {
```
YannCebron, line 62: Needing to cast is not nice. Similar code here PluginProblem.isInvalidDescriptorProblem()

novotnyr: The `ReclassifiedPluginProblem` is intentionally a subclass of a `PluginProblem`: the similar approach as in `Exception` and exception wrapping. 

In addition, reclassified plugin problems is a separate branch in inheritance hierarchy.

I don't know how to implement this without instance checking.

novotnyr: Refactored to use extension functions.

YannCebron (approved):

PR #1075, opened 2024-03-25
Distinguish errors and other plugin problems in Markdown and Console outputs

novotnyr:
In stdout output, distinguish between errors and other plugin problems:

```
The following files specified for the verification are not valid plugins:
    /opt/my-plugin.jar
        Plugin problems:
            Invalid plugin descriptor 'plugin.xml'. The plugin id should not contain the word 'intellij'.
        Additional plugin warnings:
            Invalid plugin descriptor 'plugin.xml'. Multiple dependencies (4) use the same config-file attribute value 'plugin-apexdoc-lightweight.xml': [com.jetbrains.php, com.intellij.modules.python, com.intellij.modules.ruby, com.intellij.modules.webstorm].
```

Markdown:
```
# Invalid plugin

The following file specified for the verification is not a valid plugin.

## plugin.zip

Full path: `plugin.zip`

### Plugin Problems

* Invalid plugin descriptor 'plugin.xml'. The <since-build> parameter (1) format is invalid. Ensure it is greater than <130> and represents the actual build numbers.

### Additional Plugin Warnings

* Invalid plugin descriptor 'id'. The plugin ID 'com.example.intellij' has a prefix 'com.example' that is not allowed.
````

### DIFF
--- a/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/problems/PluginProblems.kt
+++ b/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/problems/PluginProblems.kt
@@ -23,3 +23,6 @@ val PluginProblem.isInvalidDescriptorProblem: Boolean
   } else {
     this is InvalidDescriptorProblem
   }
+
+val PluginProblem.isError: Boolean
+  get() = level == PluginProblem.Level.ERROR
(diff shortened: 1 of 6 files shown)


YannCebron (approved):

hsz (approved):

novotnyr:
When tests from #1080 are merged to `master`, tests should be added as well.

LChernigovskaya (approved):

PR #1082, opened 2024-04-09
Recognize covariant return types when detecting method overrides

novotnyr:
- Recognize covariant return types when detecting method overrides. For example, remember that `String foo()` might override `Object foo()` where `String extends Object`.
- Recognize synthetic [bridge methods](http://www.angelikalanger.com/GenericsFAQ/FAQSections/TechnicalDetails.html#FAQ103) when enumerating class methods. Bridge method is allowed to call an `@OverrideOnly` method, as it is a technical method to handle inheritance and generics.
- Reorganize method helper functions which detect overrides.
- Fix false positive in the Plugin Verifier that is connected with indexes.

See [MP-6438](https://youtrack.jetbrains.com/issue/MP-6438)

### DIFF
--- a/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodsTest.kt
+++ b/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodsTest.kt
@@ -111,7 +111,7 @@ class MethodsTest {
     val arrayListToString = resolver.findMethodInClass(abstractCollectionBinaryName, toStringDesc)
     val objectToString = resolver.findMethodInClass(objectBinaryName, toStringDesc)
 
-    assertTrue(arrayListToString.doesOverride(objectToString, resolver))
+    assertTrue(arrayListToString.isOverriding(objectToString, resolver))
   }
 
   @Test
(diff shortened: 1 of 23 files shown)


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodsTest.kt
```diff
@@ -37,6 +39,145 @@ class MethodsTest {
...
+    val resolver = createTestResolver()
+    val arrayListToString = resolver.findMethodInClass(childName, getIndexerInChild)
+    val objectToString = resolver.findMethodInClass(parentName, getIndexerInParent)
+
+    assertTrue(arrayListToString.isOverriding(objectToString, resolver))
+  }
+
+
+  /*
+  @Test
```
YannCebron, line 136: ???

novotnyr: Oh, this was a very unfortunate leftover. Rewrote the unit test to have proper and matching variable names.

YannCebron (approved):

jkischkel:
Nice fix :+1:

Is there a release scheduled for this? My plugin is currently failing verification because of the false positives for `override-only`.

novotnyr:
This has been released in [https://github.com/JetBrains/intellij-plugin-verifier/releases/tag/1.367](1.367).

PR #1086, opened 2024-04-23
Refactor optional dependency parsing

novotnyr:
Plugin parsing means lots of independent activities that are intertwined and hard to extend.

- Move optional dependency parsing away from `PluginCreator` to a separate class. Simplify code and establish cleaner circular dependency tracking with `OptionalDependencyResolver` class.
- Establish an interface `PluginLoader` that indicates a class that is responsible for plugin loading and resolution. This includes transitive dependency resolution.
- Make plugin returned from the `PluginCreationResult` immutable and a clone of the plugin that is under construction by the creator. This prevents hard-to-track errors with pointer chasing where such currently constructed plugin is modified by external resolution of the dependencies. In addition, it fixes multiple issues with the propagation of structural `PluginProblem`s to the Outputs.
- Internal classes should track successfully constructed plugin by calling `isSuccess`, not by getting the `PluginCreationResult`.


intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/mocks/MockPluginsTest.kt
```diff
@@ -311,6 +321,33 @@ class MockPluginsTest(fileSystemType: FileSystemType) : BasePluginManagerTest<Id
...
+            perfectXmlBuilder.modify {
+              depends += """<depends optional="true" config-file="optionalDependency.xml">Optional Dependency</depends>"""
+            }
+          }
+
+          file(
+            "optionalDependency.xml",
+            """
+                <idea-plugin>
+                  <id>mock.OptionalDependency</id>
```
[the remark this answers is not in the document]
novotnyr: Removed `id`. This is a different scenario out of scope of this Pull Request. A good idea to validate, though.

YannCebron (approved):

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/OptionalDependencyResolver.kt
```diff
@@ -0,0 +1,62 @@
...
+    resourceResolver: ResourceResolver,
+    problemResolver: PluginCreationResultResolver,
+    dependencyChain: DependencyChain
+  ) {
+    if (!dependencyChain.extend(this)) {
+      return
+    }
+    for ((pluginDependency, configurationFile) in optionalDependenciesConfigFiles) {
+      if (dependencyChain.detectCycle(configurationFile)) {
+        return
```
LChernigovskaya, line 38: Shouldn't be there `continue` instead of `return` to gather all cycles?

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
```diff
@@ -58,6 +58,10 @@ class LevelRemappingPluginCreationResultResolver(private val delegatedResolver:
   }
 
   private fun remapPluginProblemLevel(pluginProblem: PluginProblem): PluginProblem? {
+    return remapPluginProblemLevel(pluginProblem, unwrapRemappedProblems)
+  }
+
+  private fun remapPluginProblemLevel(pluginProblem: PluginProblem, unwrapRemappedProblems: Boolean): PluginProblem? {
```
LChernigovskaya, line 64: can't understand why you need this

novotnyr:
Please note that the `clone()` is established to maintain an truly immutable copy of plugin that has been properly constructed. This means that `pluginCreationResult` property will always return an independent and an immutable instance of `IdePlugin`.

This PR addresses an issue when a plugin instance that is being created might be modified by the recursive steps of dependent plugin creations _and_ being accessed by the `PluginCreator` clients. Such entanglement will be resolved by using immutable clones.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/OptionalDependencyResolver.kt
novotnyr: Replaced with `continue` and added more unit tests. In addition, fixed a corner-case issue.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/LevelRemappingPluginCreationResultResolver.kt
novotnyr: There might be a chain of multiple level remapping resolvers. The majority of instances do not need to remap an already remapped problem, but there are cases when this is necessary.

YannCebron (approved):

LChernigovskaya (approved):

PR #1087, opened 2024-04-24
Plugin ID and plugin name property fixes

novotnyr:
- Fix unit tests for `ForbiddenPluginIdPrefix` which confuse plugin descriptor with plugin ID
- Pass plugin ID / plugin name in the plugin problem with disallowed words
- Improve copywriting


YannCebron:
regarding wording of warning/error messages: we should strive for clear use of _should_ vs _must_, e.g., by following https://www.rfc-editor.org/rfc/rfc2119

YannCebron (approved):

LChernigovskaya (approved):

PR #1101, opened 2024-06-20
Report usages of Kotlin classes with `internal` visibility modifier

novotnyr:
Report usages of `internal` visibility modifier for usages of Kotlin classes, methods and fields.

- Resolve `internal` modifier in the bytecode via `kotlinx-metadata-jvm` library
- Introduce special hierarchy of Internal API Usages for three Kotlin cases: `internal` classes, `internal` functions and `internal` field accesses.
- Extract common logic that detects internal usages, with specific scenarios for Kotlin `internal` modifier while remaining original logic for Internal API annotations.

See [MP-6650](https://youtrack.jetbrains.com/issue/MP-6650/Report-usages-of-Kotlin-classes-with-modifier-internal-in-Plugin-Verifier).


YannCebron:
I'd suggest to change message "is not supposed to be used in client code" to something mentioning usage _outside_ of the defining module.

YannCebron (approved):

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
```diff
@@ -0,0 +1,71 @@
...
+import kotlinx.metadata.jvm.KotlinClassMetadata
+import kotlinx.metadata.jvm.Metadata
+import org.objectweb.asm.tree.AnnotationNode
+import org.objectweb.asm.tree.ClassNode
+import java.util.*
+
+internal const val KOTLIN_METADATA_ANNOTATION_DESC = "Lkotlin/Metadata;"
+
+class KtClassResolver {
+  private val cache = TreeMap<ClassNode, KtClassNode>(Comparator<ClassNode> { o1, o2 ->
```
chashnikov, line 12: Why do you need to use `TreeMap` here, why isn't it enough to use a `HashMap`?

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
```diff
@@ -0,0 +1,71 @@
...
+    val s1 = o1?.signature ?: ""
+    val s2 = o2?.signature ?: ""
+    s1.compareTo(s2)
+  })
+
+  operator fun get(classNode: ClassNode): KtClassNode? {
+    return cache.getOrPut(classNode) {
+      return findMetadataAnnotation(classNode)?.let {
```
chashnikov, line 20: Please note that unqualified `return` statement in lambda expressions passed to inline function returns from the outer functions (see [the doc](https://kotlinlang.org/docs/returns.html#return-to-labels)). So actually this code never puts data to the cache. In order to return value from the lambda, you may either use qualified return or simply remove `return` statement (the last expression in the lambda is returned automatically).

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/internal/kotlin/KtInternalClassUsage.kt
```diff
@@ -0,0 +1,40 @@
...
+    get() = "Usage of class having Kotlin `internal` visibility modifier."
+
+  override val shortDescription
+    get() = "Internal " + apiElement.elementType.presentableName + " ${apiElement.formatClassLocation(FULL_NAME, NO_GENERICS)} reference"
+
+  override val fullDescription: String
+    get() = buildString {
+      val element = apiElement.elementType.presentableName
+      append("Internal $element ${apiElement.formatClassLocation(FULL_NAME, WITH_GENERICS)}")
+      append(" is referenced in " + usageLocation.formatUsageLocation())
```
chashnikov, line 29: It looks strange when string concatenations are mixed with templates and explicit `append` calls. I think it would be better to use templates instead of concatenations everywhere, and maybe replace `append` calls by raw string literal.

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
novotnyr: `ClassNode` is from ASM library and it does not provide `hashCode()`/`equals()` which is required by `HashMap`.

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
novotnyr: Well spotted, thanks for insights!

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
chashnikov: I meant that you can use `HashMap<String, KtClassNode>` and use `signature` as a key like it's done now.

chashnikov (approved):

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
```diff
@@ -0,0 +1,90 @@
...
+
+internal const val KOTLIN_METADATA_ANNOTATION_DESC = "Lkotlin/Metadata;"
+
+private typealias Signature = String
+
+class KtClassResolver {
+  private val cache = HashMap<Signature, KtClassNode>()
+
+  operator fun get(classNode: ClassNode): KtClassNode? {
+    return if (classNode in cache) {
```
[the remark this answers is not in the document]
novotnyr: Removed operators and simplified according to suggestions.

phodal:
The new versions may have introduced changes that caused my plugin to break when used with kotlinx.serialization and kotlinx.coroutines. These dependencies were working fine in the previous version.

I declared these dependencies in my build.gradle.kts. You can view the configuration here: [build.gradle.kts](https://github.com/phodal/shire/blob/main/build.gradle.kts). In the previous version, everything worked fine, but now there are 171 internal API usages being flagged.

图片

novotnyr:
Usages of `internal` Kotlin modifier are disabled in [1.373](https://github.com/JetBrains/intellij-plugin-verifier/releases/tag/1.373). They will be reenabled when such cornercases will be resolved.

PR #1103, opened 2024-06-27
OverrideOnly API declared within plugin and invoked from the plugin itself should not be reported

novotnyr:
If the plugin defines an API with `@ApiStatus.OverrideOnly`, the invocations of such API elements from the plugin itself should not be reported as problems.

- Declare an `ApiUsageFilter` that allows same-module invocations


YannCebron:
It must take into account _all_ ApiStatus variants. https://youtrack.jetbrains.com/issue/MP-6680

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/filter/SameModuleUsageFilter.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import com.jetbrains.pluginverifier.verifiers.VerificationContext
+import com.jetbrains.pluginverifier.verifiers.hasAnnotation
+import com.jetbrains.pluginverifier.verifiers.resolution.BinaryClassName
+import com.jetbrains.pluginverifier.verifiers.resolution.ClassFileMember
+import com.jetbrains.pluginverifier.verifiers.resolution.Method
+import com.jetbrains.pluginverifier.verifiers.resolution.searchParentOverrides
+import org.objectweb.asm.tree.AbstractInsnNode
+
+private const val DISABLE_SAME_MODULE_INVOCATIONS: BinaryClassName = "com/jetbrains/pluginverifier/verifiers/resolution/DisableSameModuleInvocations"
```
YannCebron, line 11: It would be better to not have test-environment-only annotations present and handled in production code. Maybe some environment parameter instead set for corresponding test(s) explicitly?

novotnyr: Resolved via effective element that declares the corresponding annotation. Opt-out annotation has been removed.

novotnyr:
> It must take into account _all_ ApiStatus variants. https://youtrack.jetbrains.com/issue/MP-6680 could we please have some tests for the others as well?

Other annotations have a different mechanisms. I suggest focusing this PR on the `OverrideOnly`. Other `ApiStatus`-es will be targeted by the subsequent PR.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/DisableSameModuleInvocations.kt
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Indicates that the invocation of the target API element is prohibited
+ * even if it occurs within the same module.
+ *
+ * Mainly, this is used in tests to verify OverrideOnly invocations in the same module or plugin.
+ */
+@Retention(AnnotationRetention.RUNTIME)
+@Target(AnnotationTarget.FUNCTION)
+annotation class DisableSameModuleInvocations
```
YannCebron, line 11: remove now?

novotnyr: Removed.

YannCebron (approved):

LChernigovskaya (approved):

PR #1105, opened 2024-07-01
Check for `ApiStatus` annotations that occur within the same module

novotnyr:
If the plugin defines an API with `@ApiStatus` annotations, the invocations of such API elements from the plugin itself should not be reported as problems.

This is an extension of #1103.

The following annotations are supported:

- `@ApiStatus.Experimental`
- `@ApiStatus.OverrideOnly` (as per #1103)
- `@ApiStatus.ScheduledForRemoval`
- `@ApiStatus.Internal`

The following annotations are not supported
- The `@ApiStatus.NonExtendable`: the Plugin Verifier does not track such internal Platform calls. However, the annotation is tracked as a special case of API Usage.
- `@ApiStatus.AvailableSince`: not supported.
- `@ApiStatus.Obsolete`: not supported in the Plugin Verifier.


YannCebron (approved):

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/verifiers/PluginVerificationContext.kt
```diff
@@ -103,11 +102,7 @@ data class PluginVerificationContext(
...
   override fun registerDeprecatedUsage(deprecatedApiUsage: DeprecatedApiUsage) {
-    val deprecatedElementHost = deprecatedApiUsage.apiElement.containingClass
-    val usageHostClass = deprecatedApiUsage.usageLocation.containingClass
-    if (deprecatedApiUsage is DiscouragingJdkClassUsage || shouldIndexDeprecatedClass(usageHostClass, deprecatedElementHost)) {
-      deprecatedUsages += deprecatedApiUsage
-    }
+    deprecatedUsages += deprecatedApiUsage
   }
 
   override fun registerExperimentalApiUsage(experimentalApiUsage: ExperimentalApiUsage) {
```
LChernigovskaya, line 108: Can't write comment for the code below for some reason. The question for `if (shouldIndexDeprecatedClass(usageHostClass, elementHostClass))` -- shouldn't it be removed as well?

LChernigovskaya:
And just to know: why can't `samePluginUsageFilter` be used for `overrideOnlyUsageFilter` (instead of `SameModuleUsageFilter`)?

novotnyr:
> And just to know: why can't `samePluginUsageFilter` be used for `overrideOnlyUsageFilter` (instead of `SameModuleUsageFilter`)?

* `SameModuleUsageFilter` is annotation-specific. It detects usages of a particular annotation within the same module.
* `SamePluginUsageFilter` is invocation-specific.

LChernigovskaya (approved):

PR #1113, opened 2024-07-09
Add unit test for `OverrideOnly` checks

novotnyr:
Add unit tests for cornercases:

- static method invocation: when a method invokes a `static` method with exactly the same name and exactly the same descriptor on another class
- `super` method invocation: when a method just plainly invokes a parent method

See #1111 and #1112

### DIFF
--- a/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/usages/overrideOnly/DelegateCallOnOverrideOnlyUsageFilterTest.kt
+++ b/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/usages/overrideOnly/DelegateCallOnOverrideOnlyUsageFilterTest.kt
@@ -2,6 +2,8 @@ package com.jetbrains.pluginverifier.usages.overrideOnly
 
 import com.jetbrains.plugin.structure.classes.resolvers.FileOrigin
 import com.jetbrains.pluginverifier.tests.mocks.MockVerificationContext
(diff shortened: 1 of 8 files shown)


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/usages/overrideOnly/DelegateCallOnOverrideOnlyUsageFilterTest.kt
```diff
@@ -0,0 +1,124 @@
...
+  fun setUp() {
+    verificationContext = MockVerificationContext()
+    filter = DelegateCallOnOverrideOnlyUsageFilter()
+  }
+
+  @Test
+  fun `invocation of super as delegate is intentionally ignored`() {
+    val className = "mock.plugin.overrideOnly.ClearCountingContainer"
+    val clearMethod = ClearCountingContainerNode().loadMethod(CLEAR_METHOD, NO_PARAMS_RETURN_VOID_DESCRIPTOR)
+    if (clearMethod == null) {
```
YannCebron, line 42: why not assertNotNull() or similar

novotnyr: Very good point, replaced in the code.

YannCebron (approved):

PR #1127, opened 2024-07-26
Match problems in the plugin and problems in the creation result

novotnyr:
- Fix issue when `PluginCreationSuccess` contains an empty set of problems (due to remappings) and the plugin itself contains the original non-remapped problems.
- When cloning plugin, decide if the provided list of problems should override or append to the list of the plugin that should be cloned.
- Add tests that emulate JetBrains Marketplace setup of plugin problem remapping for existing plugins.

See [MP-6773](https://youtrack.jetbrains.com/issue/MP-6773) Plugin Verifier: Problems in the plugin and problems in the creation result do not match

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginImpl.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginImpl.kt
@@ -84,7 +84,7 @@ class IdePluginImpl : IdePlugin, StructurallyValidated {
     (pluginId ?: pluginName ?: "<unknown plugin ID>") + (pluginVersion?.let { ":$it" } ?: "")
 
   companion object {
-    fun clone(old: IdePlugin, structureProblems: List<PluginProblem>): IdePlugin {
+    fun clone(old: IdePlugin, structureProblems: List<PluginProblem>, overrideStructureProblems: Boolean): IdePlugin {
       return IdePluginImpl().apply {
         pluginId = old.pluginId
         pluginName = old.pluginName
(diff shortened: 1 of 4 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginImpl.kt
```diff
@@ -120,10 +120,11 @@ class IdePluginImpl : IdePlugin, StructurallyValidated {
         modulesDescriptors.addAll(old.modulesDescriptors)
         thirdPartyDependencies = old.thirdPartyDependencies.toMutableList()
         if (old is StructurallyValidated) {
-          if (structureProblems.isNotEmpty()) {
+          if (overrideStructureProblems) {
             problems.addAll(structureProblems)
           } else {
-            problems.addAll(old.problems)
+            val oldProblems = (structureProblems + old.problems).toSet()
+            problems.addAll(oldProblems)
```
LChernigovskaya, line 127: Won't there be a mess that in case of `overrideStructureProblems` is false, copied plugin will contain both original and remapped problems? 

Also, probably, it would make sense to rename `overrideStructureProblems`, because now it's a little bit strange that we override structure problems by `structureProblems`.

novotnyr: Removed and simplified semantics.

The method is used in a single place with specific semantics, no need to make it more complex.

novotnyr:
Reworked semantics of the `clone` method. Since it is used in a single place, there is no need to make it more complex than necessary.

Additionally, I have added KDoc.

YannCebron (approved):

LChernigovskaya (approved):

PR #1131, opened 2024-08-05
Disable Kotlin `internal` modifier usage check

novotnyr:
Disable API usage checks of Kotlin `internal` modifier.

Rationale: The current implementation triggers many false positives, such as [MP-6784](https://youtrack.jetbrains.com/issue/MP-6784). Additional effort is required, with a scope extending beyond the 2024.2 release.

### DIFF
--- a/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/KotlinInternalModifierUsageTest.kt
+++ b/intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/KotlinInternalModifierUsageTest.kt
@@ -127,10 +127,9 @@ class KotlinInternalModifierUsageTest {
     val plugin = prepareUsage(pluginSpec) { usageClassUdt }
 
     verify(ide, plugin).run {
-      assertEquals(1, size)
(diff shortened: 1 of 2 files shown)


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/KotlinInternalModifierUsageTest.kt
```diff
@@ -127,10 +127,9 @@ class KotlinInternalModifierUsageTest {
     val plugin = prepareUsage(pluginSpec) { usageClassUdt }
 
     verify(ide, plugin).run {
-      assertEquals(1, size)
+      assertEquals(0, size)
```
YannCebron, line 130: why not simply disable/mute the test altogether?

novotnyr: Using `@Ignore` annotations instead. Restored the original behaviour.

YannCebron (approved):

mkondratek:
thanks! 🙇‍♂️

PR #1135, opened 2024-08-15
Provide a list of ignored packages when tracking Kotlin `internal` API usages

novotnyr:
Some Kotlin API usages with `internal` visibility modifier are generated by code generators.

Common examples:

- `launch{}` coroutine builder maps to `SuspendLambda` class usage
- Kotlin Serialization uses various internal classes.

## Solution 

Provide a mechanism to exclude such use-cases:

1. Do not track usages that are marked as [`@PublishedApi`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-published-api/)
2. Ignore specific packages as a hardcoded list when `@PublishedApi` is not available.
  -  `kotlin.jvm.internal`,
  -  `kotlin.internal`,
  -  `kotlin.coroutines.jvm.internal`,
  -  `kotlinx.serialization.internal`
3. Add a Gradle setup that allows to test against real Kotlin Stdlib in mock IDEs.

## Technical aspects

- Support `@PublishedApi` in the `KtInternalModifierUsageProcessor`
- Introduce `KtInternalModifierUsageFilter` and plug it into `PluginVerifier` to exclude specific packages from problem reporting.

## Asides

Tests have been ran against large-scale set of plugins in the JetBrains Marketplace. There are few detected corner-cases beyond supported packages.

- `kotlin.random.AbstractPlatformRandom` used by 3 outdated plugins that are more than 3 years old.
- `kotlin.sequences.Sequence`: used by single outdated plugin
- `kotlin.coroutines.Continuation`: used by two outdated plugins
- `kotlin.Result.Failure`: used by 4 outdated plugins
- `kotlinx.coroutines.channels.ChannelResult.Closed` and `kotlinx.coroutines.channels.ChannelResult.Failed`: used by two JetBrains plugins.

See [MP-6784](https://youtrack.jetbrains.com/issue/MP-6784) and [MP-6911](https://youtrack.jetbrains.com/issue/MP-6911).


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/Dumps.kt
```diff
@@ -0,0 +1,342 @@
...
+import org.objectweb.asm.AnnotationVisitor
+import org.objectweb.asm.ClassWriter
+import org.objectweb.asm.FieldVisitor
+import org.objectweb.asm.Label
+import org.objectweb.asm.MethodVisitor
+import org.objectweb.asm.Opcodes
+import org.objectweb.asm.Opcodes.*
+
+
+object Dumps {
```
YannCebron, line 12: add test sources and compiled`.jar` file to testdata folder containing relevant cases instead of manually outputting?

YannCebron:
What about the cases listed in "Aside", should they be added to hardcoded exceptions as well?

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/Dumps.kt
YannCebron: actually standalone bytecode is easier to maintain

YannCebron (approved):

novotnyr:
The corner-cases will be investigated in [MP-6911](https://youtrack.jetbrains.com/issue/MP-6911).

PR #1136, opened 2024-08-16
IDE Diff Builder: Migrate API usages in the structure library

novotnyr:
In `ide-diff-builder` migrate API usages in the `structure` library to the recent API.


hsz (approved):
LGTM

PR #1140, opened 2024-08-21
Validate 'release-version' for paid plugins

novotnyr:
[JetBrains Marketplace Guidelines](https://plugins.jetbrains.com/docs/marketplace/add-required-parameters.html) specify validation rules for `release-version` parameter that are applied on paid plugins.

New Plugin Problems
-------------------

1. `ReleaseVersionWrongFormat` when a `release-version` does not have two digits. This is an _error_ for new plugins and a _warning_ for existing plugins.
2. `ReleaseVersionAndPluginVersionMismatch` when a `release-version` does match the plugin version. This is an _error_ for both new plugins and existing plugins, and a _ignored_ problem for JetBrains plugins.

Technical improvements
----------------------

- Introduce `ProductReleaseVersionVerifier` as a separate class for `product-release` verification
- Introduce type-safe `ProductReleaseVersion` that holds major and minor components of product release number.
- Deprecate`releaseVersion` field in `ProductDescriptor` in favour of type-safe`version`.

Level remapping notes
----------------------
`ReleaseVersionAndPluginVersionMismatch` is at `ERROR` level for both existing and new plugins.
Some plugins were early adopters of the paid plugins feature, but their versions might not meet the current requirements.
Such plugins are explicitly handled on the JetBrains Marketplace side.

Such error can be muted via `-mute` CLI option.

See [MP-6824](https://youtrack.jetbrains.com/issue/MP-6824) Plugin Verifier: Improve validation of `release-version`

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
@@ -226,6 +226,12 @@ class ReleaseVersionAndPluginVersionMismatch(
 ) {
   override val level
     get() = Level.ERROR
+
+  override val hint: ProblemSolutionHint
+    get() = ProblemSolutionHint(
+      "release-version=\"20201\" and <version>2020.1</version>",
+      "https://plugins.jetbrains.com/docs/marketplace/add-required-parameters.html"
+    )
 }
 
 class UnableToFindTheme(descriptorPath: String, themePath: String) : InvalidDescriptorProblem(
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2000-2020 JetBrains s.r.o. and other contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
+ * Copyright 2000-2024 JetBrains s.r.o. and other contributors. Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE file.
  */
 
 package com.jetbrains.plugin.structure.intellij.problems
@@ -7,6 +7,7 @@ package com.jetbrains.plugin.structure.intellij.problems
 import com.jetbrains.plugin.structure.base.problems.InvalidDescriptorProblem
 import com.jetbrains.plugin.structure.base.problems.ProblemSolutionHint
 import com.jetbrains.plugin.structure.intellij.version.IdeVersion
+import com.jetbrains.plugin.structure.intellij.version.ProductReleaseVersion
 
 class PropertyWithDefaultValue(
   descriptorPath: String,
@@ -208,6 +209,20 @@ class ReleaseVersionWrongFormat(
   descriptorPath = descriptorPath,
   detailedMessage = "The <release-version> parameter ($releaseVersion) format is invalid. " +
     "Ensure it is an integer with at least two digits."
+) {
+  override val level
+    get() = Level.ERROR
+}
+
+class ReleaseVersionAndPluginVersionMismatch(
+  descriptorPath: String,
+  releaseVersion: ProductReleaseVersion,
+  pluginVersion: String
+) : InvalidDescriptorProblem(
+  descriptorPath = descriptorPath,
+  detailedMessage = "The <release-version> parameter [$releaseVersion] and the plugin version [$pluginVersion] " +
+    "should have a matching beginning. " +
+    "For example, release version '20201' should match plugin version 2020.1.1"
 ) {
   override val level
     get() = Level.ERROR
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
@@ -201,6 +201,18 @@ class ReleaseDateInFuture(descriptorPath: String) : InvalidDescriptorProblem(
     get() = Level.ERROR
 }
 
+class ReleaseVersionWrongFormat(
+  descriptorPath: String,
+  releaseVersion: String
+) : InvalidDescriptorProblem(
+  descriptorPath = descriptorPath,
+  detailedMessage = "The <release-version> parameter ($releaseVersion) format is invalid. " +
+    "Ensure it is an integer with at least two digits."
+) {
+  override val level
+    get() = Level.ERROR
+}
+
 class UnableToFindTheme(descriptorPath: String, themePath: String) : InvalidDescriptorProblem(
   descriptorPath = descriptorPath,
   detailedMessage = "The theme description file cannot be found by the path '$themePath'. Ensure the theme description " +
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/MinorWarnings.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/MinorWarnings.kt
@@ -9,7 +9,6 @@ import com.jetbrains.plugin.structure.base.problems.PluginDescriptorResolutionEr
 import com.jetbrains.plugin.structure.base.problems.PluginProblem
 import com.jetbrains.plugin.structure.base.problems.ProblemSolutionHint
 import com.jetbrains.plugin.structure.intellij.version.IdeVersion
-import com.jetbrains.plugin.structure.intellij.version.ProductReleaseVersion
 
 class NoModuleDependencies(descriptorPath: String) : InvalidDescriptorProblem(
   descriptorPath = descriptorPath,
@@ -179,20 +178,6 @@ open class NonexistentReleaseInUntilBuild(
   untilBuild: String,
   nonexistentRelease: String = ""
 ) : SuspiciousUntilBuild(untilBuild, "Version '$nonexistentRelease' does not exist")
-
-class SuspiciousReleaseVersion(
-  descriptorPath: String,
-  releaseVersion: ProductReleaseVersion,
-  pluginVersion: String
-) : InvalidDescriptorProblem(
-  descriptorPath = descriptorPath,
-  detailedMessage = "The <release-version> parameter [$releaseVersion] and the plugin version [$pluginVersion] " +
-    "should have a matching beginning. " +
-    "For example, release version '20201' should match plugin version 2020.1.1"
-) {
-  override val level
-    get() = Level.WARNING
-}
 
 class ForbiddenPluginIdPrefix(
   descriptorPath: String,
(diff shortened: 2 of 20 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/version/ProductReleaseVersion.kt
```diff
@@ -0,0 +1,36 @@
+package com.jetbrains.plugin.structure.intellij.version
+
+
+private const val RELEASE_VERSION_TAG_NAME = "release-version"
```
karollewandowski, line 4: Unused? If it is used implicitly, there is no `release-version` tag.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/version/ProductReleaseVersion.kt
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Type-safe `release-version` of paid plugins.
+ *
+ * Relevant docs:
+ * - See
+ * [Plugin Configuration file](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#idea-plugin__product-descriptor)
+ * at IntelliJ Platform Plugin SDK
+ * - [JetBrains Marketplace](https://plugins.jetbrains.com/docs/marketplace/add-required-parameters.html) guidelines
+ */
+data class ProductReleaseVersion(val value: Int) {
```
karollewandowski, line 14: I understand it handles versions with a single-digit minor version.

Are we sure there are no plugins releasing a major version more than 9 times a year? For example, `202412`. I think it is worth checking if it was not.

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/intellij/PluginXmlValidationTest.kt
```diff
@@ -27,6 +31,15 @@ private const val HEADER = """
       <idea-version since-build="131.1"/>
     """
 
+private const val HEADER_WITHOUT_VERSION = """
+      <id>someId</id>
+      <name>someName</name>
+      ""<vendor email="vendor.com" url="url">vendor</vendor>""
```
karollewandowski, line 37: What are these empty strings at the beginning and end of the line for?

intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/intellij/PluginXmlValidationTest.kt
```diff
@@ -130,6 +143,82 @@ class PluginXmlValidationTest {
...
+    assertNotNull(plugin.productDescriptor)
+    with(plugin.productDescriptor!!) {
+      assertEquals("PCODE", code)
+      assertEquals(LocalDate.of(2024, 8, 13), releaseDate)
+      assertEquals(ProductReleaseVersion(10), version)
+    }
+  }
+
+  @Test
+  fun `paid plugin with a incorrect single-digit release version`() {
```
karollewandowski, line 172:
```suggestion
  fun `paid plugin with an incorrect single-digit release version`() {
```

karollewandowski (changes requested):

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
```diff
@@ -201,6 +201,18 @@ class ReleaseDateInFuture(descriptorPath: String) : InvalidDescriptorProblem(
     get() = Level.ERROR
 }
 
+class ReleaseVersionWrongFormat(
+  descriptorPath: String,
+  releaseVersion: String
+) : InvalidDescriptorProblem(
+  descriptorPath = descriptorPath,
+  detailedMessage = "The <release-version> parameter ($releaseVersion) format is invalid. " +
```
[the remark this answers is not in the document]
novotnyr: This is a project convention. We do not distinguis between attributes and elements. Everything is - historically - indicated as a _parameter_ with `<`..`>` as delimiters.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/version/ProductReleaseVersion.kt
```diff
@@ -0,0 +1,36 @@
...
+  val major: Int
+    get() = if (isSingleDigit) value else value / 10
+
+  val minor: Int
+    get() = if (isSingleDigit) 0 else value % 10
+
+  val isSingleDigit: Boolean
+    get() = value in 1..9
```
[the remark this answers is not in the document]
novotnyr: The single-digit release version is used by the early adopters of paid plugin scheme. There is no plugin with zero release version in the JetBrains Marketplace.

karollewandowski (approved):
Looks good to me.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/version/ProductReleaseVersion.kt
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * Relevant docs:
+ * - See
+ * [Plugin Configuration file](https://plugins.jetbrains.com/docs/intellij/plugin-configuration-file.html#idea-plugin__product-descriptor)
+ * at IntelliJ Platform Plugin SDK
+ * - [JetBrains Marketplace](https://plugins.jetbrains.com/docs/marketplace/add-required-parameters.html) guidelines
+ */
+data class ProductReleaseVersion(val value: Int) {
+  val major: Int
+    get() = if (isSingleDigit) value else value / 10
```
LChernigovskaya, line 15: if release-version is a single digit, the `major` is 0 and the `minor` is value

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/verifiers/ProductReleaseVersionVerifier.kt
```diff
@@ -0,0 +1,67 @@
...
+
+  private fun verifyPluginVersionAndReleaseVersionMatch(
+    plugin: PluginBean,
+    productReleaseVersion: ProductReleaseVersion,
+    descriptorPath: String,
+    problemRegistrar: ProblemRegistrar
+  ) {
+    if (plugin.pluginVersion == null) return
+
+    val commonPrefix = productReleaseVersion.major.toString().commonPrefixWith(plugin.pluginVersion)
```
LChernigovskaya, line 48: Shouldn't be `minor` component also be checked. If I'm not mistaken it allows to have releaseVersion = 20221 and version = 2022.5.2?

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/MinorWarnings.kt
```diff
@@ -179,6 +180,19 @@ open class NonexistentReleaseInUntilBuild(
...
+  releaseVersion: ProductReleaseVersion,
+  pluginVersion: String
+) : InvalidDescriptorProblem(
+  descriptorPath = descriptorPath,
+  detailedMessage = "The <release-version> parameter [$releaseVersion] and the plugin version [$pluginVersion] " +
+    "should have a matching beginning. " +
+    "For example, release version '20201' should match plugin version 2020.1.1"
+) {
+  override val level
+    get() = Level.WARNING
```
LChernigovskaya, line 194: We don't allow upload with such problem. Probably it should be error for new, warning for existing ones and ignored for JetBrains (since JetBrains plugins have very strange release versions like `2024200` instead of `20242`)

LChernigovskaya (changes requested):
Could you please tell about the purpose of having these checks in structure also? Unfortunately, it won't allow to get rid of these checks in Marketplace. At least from checking the matching of version and release-version, since we need to check it for each updates not only for the first one. Probably, we can pass a new parameter telling if this plugin should be checked for correct paid information or not (like `new`/`existing` parameter), so all our checks might be transferred to structure, since it'll be possible to not make some checks for old paid plugins. In current state (without this new parameter) I see that there will be 2 the same errors during upload -- one from structure check, another one from Marketplace.

novotnyr:
> Could you please tell about the purpose of having these checks in structure also? Unfortunately, it won't allow to get rid of these checks in Marketplace. At least from checking the matching of version and release-version, since we need to check it for each updates not only for the first one. Probably, we can pass a new parameter telling if this plugin should be checked for correct paid information or not (like `new`/`existing` parameter), so all our checks might be transferred to structure, since it'll be possible to not make some checks for old paid plugins. In current state (without this new parameter) I see that there will be 2 the same errors during upload -- one from structure check, another one from Marketplace.

Plugin authors will be notified about issues in the paid plugins which can they be aware of and fix.

As for JetBrains Marketplace, explicit problems can be muted in the API by wrapping an existing resolver with `LevelRemappingPluginCreationResultResolver`. See [the corresponding unit test for usage](https://github.com/JetBrains/intellij-plugin-verifier/pull/1140/commits/3da572da2a6920dea1985336ae5c407083c0ddf2).

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/verifiers/ProductReleaseVersionVerifier.kt
```diff
@@ -0,0 +1,90 @@
...
+    if (releaseVersionValue.isNullOrEmpty()) {
+      return Invalid("Attribute '$RELEASE_VERSION_ATTRIBUTE_NAME' is missing").also {
+        problemRegistrar.registerProblem(PropertyNotSpecified(RELEASE_VERSION_ATTRIBUTE_NAME, descriptorPath))
+      }
+    }
+
+    return try {
+      ProductReleaseVersion.parse(releaseVersionValue).run {
+        if (isSingleDigit) {
+          Invalid("Attribute '$RELEASE_VERSION_ATTRIBUTE_NAME' must have two or more digits: '$releaseVersionValue'").also {
```
LChernigovskaya, line 25: Will it be possible to remap this to `warning` for old plugins, which still have one digit in the release-version?

intellij-plugin-structure/structure-intellij/src/main/resources/com/jetbrains/plugin/structure/intellij/problems/plugin-problems.json
```diff
@@ -4,7 +4,9 @@
     "com.jetbrains.plugin.structure.intellij.problems.TemplateWordInPluginId": "ignore",
     "com.jetbrains.plugin.structure.intellij.problems.TemplateWordInPluginName": "ignore",
     "com.jetbrains.plugin.structure.intellij.problems.InvalidUntilBuildWithJustBranch": "unacceptable-warning",
-    "com.jetbrains.plugin.structure.intellij.problems.InvalidUntilBuildWithMagicNumber": "unacceptable-warning"
+    "com.jetbrains.plugin.structure.intellij.problems.InvalidUntilBuildWithMagicNumber": "unacceptable-warning",
+    "com.jetbrains.plugin.structure.intellij.problems.ReleaseVersionWrongFormat": "warning",
+    "com.jetbrains.plugin.structure.intellij.problems.ReleaseVersionAndPluginVersionMismatch": "warning"
```
LChernigovskaya, line 9: Let's have `ReleaseVersionAndPluginVersionMismatch` as an `error` for existing plugins also, and then we'll remap it on the Marketplace side?

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/ExistingPluginValidationTest.kt
```diff
@@ -507,6 +508,46 @@ class ExistingPluginValidationTest : BasePluginTest() {
...
+          """
+        }
+      }
+    }
+    assertSuccess(result)
+    assertMatchingPluginProblems(result as PluginCreationSuccess)
+  }
+
+  @Test
+  fun `existing paid plugin has release-version that does not match plugin version but this is filtered`() {
```
LChernigovskaya, line 532: Probably, it's needed to change the test name

PR #1144, opened 2024-08-26
Support resource bundle `@PropertyKey` in constructors of enum classes

novotnyr:
Support resource bundle `@PropertyKey` in constructors of enum classes, such as

```kotlin
enum class ExternalLinter(@PropertyKey(resourceBundle = BUNDLE) val titleKey: String) {
    CARGO_CHECK("rust.external.linter.cargo.check.item"),
    CLIPPY("rust.external.linter.clippy.item");
}
```

## Idea

1. For each API method usage, we detect if such method is in an `enum class` and if this is a constructor invocation. As a side note, `enum classes` do not support inheritance. This happens via `ApiUsageProcessor` mechanism.
2. Via `kotlinx.metadata` mechanism, we discover the `@PropertyKey` annotation and its `resourceBundle` attribute value.
4. Via ASM `Analyzer` mechanism and custom pluggable type-safe `Interpreter`, we verify if such annotated method invokes a parent quasi-synthetic private constructor generated by the Kotlin compiler. If so, we extract the name of the resource bundle key from the byte-code by inspecting frame stack.
5. Then, we check the existence of the resource bundle key in the corresponding resource bundle.

## Technical details and changes

- Detect invocations of private enum constructors generated by Kotlin compiler. Each `enum class` gets an automatic private constructor with additional parameters: `String` as an enum member name, `Int` as an enum ordinal. Such parameters are compiled _before_ additional explicit parameters provided in the constructor. Taking the example from above, the `ExternalLinter` byte-code will contain an `private constructor(v: String, ordinal: Int, titleKey: String)` generated by the Kotlin compiler.
- Introduce an type-safe adapter for ASM `Analyzer` that is able to observe specific instructions:
  * single-integer pushing instructions
  * `INVOKESPECIAL`
- Detect private enum constructor invocations with ASM analyzer
- Extract away resource bundle property key usages into parent class.
- Establish a specific resource bundle property key usage detector `EnumPropertyUsageProcessor` that is pluggable into start resource bundle property key usage detection
- Add multiple helpers that allow to track and debug ASM instructions
 

See [MP-6710](https://youtrack.jetbrains.com/issue/MP-6710) Plugin verifier false positive "Reference to a missing property  of resource bundle"


intellij-plugin-verifier/verifier-test/mock-plugin/src/main/resources/mock/plugin/property/linter.properties
```diff
@@ -0,0 +1,2 @@
+mock.plugin.linter.quiet=Quiet Linter
+mock.plugin.linter.verbose=verbose Linter
```
karollewandowski, line 2:
```suggestion
mock.plugin.linter.verbose=Verbose Linter
```
I know it is insignificant, but still 🙃

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/ByteCodeDescriptions.kt
```diff
@@ -0,0 +1,127 @@
+package com.jetbrains.pluginverifier.tests.bytecode
+
+import org.objectweb.asm.Opcodes.*
+import org.objectweb.asm.tree.InsnNode
+import org.objectweb.asm.tree.TypeInsnNode
+
+@Suppress("unused")
+fun TypeInsnNode.toDebugString(): String = when (opcode) {
```
karollewandowski, line 8: Why `toDebugString()` are needed? Are they used implicitly? Maybe it would be good to explain it above "unused" suppressions.

karollewandowski (approved):
I didn't spot any issues, but keep in mind I don't know the project and my review was rather shallow.

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassNode.kt
```diff
@@ -9,11 +9,14 @@ import org.objectweb.asm.tree.ClassNode
...
+  val name: String
+    get() = classNode.name
+
   val isInternal: Boolean
     get() = cls.visibility == Visibility.INTERNAL
 
-  val name: String
-    get() = classNode.name
+  val isEnumClass: Boolean
+    get() = cls.enumEntries.isNotEmpty()
```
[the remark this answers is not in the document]
novotnyr: Thanks for the hint, your idea is much better approach.

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/ByteCodeDescriptions.kt
novotnyr: Added a KDoc to indicate that this is used in IDE debugging sessions.

PR #1146, opened 2024-08-27
Allow invocation of `private` interface methods when in `INVOKEDYNAMIC` calls

novotnyr:
Some Java compilers might emit `INVOKEINTERFACE` calls on private interface methods when using `INVOKEDYNAMIC` instruction. Usually, this happens in lambdas, which are compiled as `private synthetic` methods.

Allow such usages to prevent false positives.

An artificial example with intentionally simplified usage:

```java
public interface Handler {
    default void handle() {
        Supplier<Integer> s = () -> getValue();
    }

    default int getValue() {
        return 0;
    }
}
```

See [MP-6845](https://youtrack.jetbrains.com/issue/MP-6845).


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/usages/overrideOnly/Dumps.kt
```diff
@@ -213,4 +217,99 @@ private fun dumpPackageInvokingBox(): ByteArray {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Simulates a situation when Java 17 builds a class which
+ * uses `INVOKEINTERFACE` instruction on a lambda.
+ *
+ * The lambda compiles to a `private default synthetic` method,
+ * and it is dynamically dispatch via `INVOKEDYNAMIC`.
```
karollewandowski, line 227:
```suggestion
 * and it is dynamically dispatched via `INVOKEDYNAMIC`.
```

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/instruction/MethodInvokeInstructionVerifier.kt
```diff
@@ -128,10 +133,18 @@ class MethodInvokeInstructionVerifier(
       method.name == System.getProperty("plugin.verifier.test.private.interface.method.name")
 
     /*
-    Otherwise, if the resolved method is static or private, the invokeinterface instruction throws an IncompatibleClassChangeError.
+    Otherwise, if the resolved method is static or private, the INVOKEINTERFACE instruction throws an IncompatibleClassChangeError.
+    However, Java 17 and other compilers might invoke such methods when using INVOKEDYNAMIC.
+    Usuallly, this happens when invoking lambdas that are compiled as `private synthetic` methods.
```
karollewandowski, line 138:
```suggestion
    Usually, this happens when invoking lambdas that are compiled as `private synthetic` methods.
```

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/BinaryClassContent.kt
```diff
@@ -0,0 +1,23 @@
+package com.jetbrains.pluginverifier.tests.bytecode
+
+import com.jetbrains.pluginverifier.verifiers.resolution.BinaryClassName
+
+internal data class BinaryClassContent(val name: BinaryClassName, val bytes: ByteArray) {
+  override fun equals(other: Any?): Boolean {
```
karollewandowski, line 6: Why are `equals` and `hashCode` implemented explicitly? Kotlin data classes provide them out-of-the-box. Aren't they sufficient?

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/mocks/PluginBuilders.kt
```diff
@@ -0,0 +1,55 @@
...
+
+internal fun Path.buildIdePlugin(pluginContentBuilder: (ContentBuilder).() -> Unit): IdePlugin {
+  val pluginFile = buildZipFile(zipFile = this) {
+    pluginContentBuilder()
+  }
+  return (IdePluginManager.createManager().createPlugin(pluginFile) as PluginCreationSuccess).plugin
+}
+
+/**
+ * Build a bare-bones IDE without any plugins except the core plugin.
```
karollewandowski, line 24:
```suggestion
 * Builds a bare-bones IDE without any plugins except the core plugin.
```

karollewandowski (approved):

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/bytecode/BinaryClassContent.kt
novotnyr: `ByteArray` does not provide `equals()`/`hashCode()`. In addition, Kotlin/IDEA suggests that array-based properties in data classes should provide these methods explicitly.

PR #1149, opened 2024-09-04
Tc actions yaml

kesarevs:
As a part of [MP-6835](https://youtrack.jetbrains.com/issue/MP-6835/Add-a-way-to-fetch-TeamCity-Action-yaml-file) we need a way to get the content of yaml file for parsed TC Actions.

### DIFF
--- a/intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/ReSharperPluginManager.kt
+++ b/intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/ReSharperPluginManager.kt
@@ -116,6 +116,7 @@ class ReSharperPluginManager private constructor(private val extractDirectory: P
       idParts.size > 1 -> idParts[1]
       else -> id
     }
+    val nuspecFileName = "$id.nuspec"
     ReSharperPlugin(
       pluginId = id,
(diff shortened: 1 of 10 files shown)


intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/ReSharperPluginManager.kt
```diff
@@ -116,6 +116,7 @@ class ReSharperPluginManager private constructor(private val extractDirectory: P
       idParts.size > 1 -> idParts[1]
       else -> id
     }
+    val nuspecFileName = "$id.nuspec"
```
LChernigovskaya, line 119: Shouldn't we get the real name here also (from `descriptorFile.fileName`)?

LChernigovskaya (approved):

intellij-plugin-structure/structure-dotnet/src/main/kotlin/com/jetbrains/plugin/structure/dotnet/ReSharperPluginManager.kt
kesarevs: Well, we will rename it in structure to "$id.nuspec", so...

PR #1157, opened 2024-09-19
Don't use JetBrainsPluginCreationResultResolver as a default resolver

LChernigovskaya:


novotnyr:
What is the rationale behind this change?

novotnyr:
This PR makes problem remapping consistent between JetBrains Marketplace and standalone verifications.

novotnyr (approved):

PR #1161, opened 2024-09-24
Remove JetBrains vendor check from `PluginIdVerifier`

LChernigovskaya:
Remove JetBrains vendor check from `PluginIdVerifier`.

This is a duplicate logic. Any `TemplateWordInPluginId` and `ForbiddenPluginIdPrefix` that are emitted from the `PluginIdVerifier` are already suppresed by the plugin problem remapping mechanism.

Any client library can customize this behaviour by using a proper `PluginCreationResultResolver`. For example, a `JetBrainsPluginCreationResultResolver` can wrap a `LevelRemappingPluginCreationResultResolver`. The latter can explicitly suppres both of these plugin problems.

### DIFF
--- a/intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/intellij/verifiers/PluginIdVerifierTest.kt
+++ b/intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/intellij/verifiers/PluginIdVerifierTest.kt
@@ -26,18 +26,6 @@ class PluginIdVerifierTest {
   }
 
   @Test
-  fun `plugin by JetBrains has no issues`() {
-    val comIntellijPlugin = plugin("com.intellij", "JetBrains")
-    val ideaCorePlugin = plugin("IDEA CORE", "JetBrains")
-
-    verifier.verify(comIntellijPlugin, DESCRIPTOR_PATH, problemRegistrar)
-    verifier.verify(ideaCorePlugin, DESCRIPTOR_PATH, problemRegistrar)
-
-    Assert.assertTrue(problems.isEmpty())
-  }
-
-
-  @Test
   fun `plugin by 3rd party is disallowed`() {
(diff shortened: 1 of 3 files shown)


intellij-plugin-structure/tests/src/test/kotlin/com/jetbrains/plugin/structure/intellij/verifiers/PluginIdVerifierTest.kt
```diff
@@ -25,18 +25,6 @@ class PluginIdVerifierTest {
     problems = mutableListOf()
   }
 
-  @Test
```
LChernigovskaya, line 28: There is a similar test in `JetBrainsPluginTest`

YannCebron:
What is the purpose of this removal? How will this affect existing plugins by JB?

LChernigovskaya:
> What is the purpose of this removal? How will this affect existing plugins by JB?

It's connected with recent changes, that the information if the plugin is JetBrains or not will be passed from Marketplace. But for CLI there will be the old solution when the vendor from XML is checked.
I found that there is also this check (probably it was added before remapping), so with this code we check that the vendor is JetBrains twice, first here and ignore this error, then during remapping. It should be enough to check it only during remapping.

novotnyr:
`PluginIdVerifier` uses a shortcut that ignores ID verification rules for JetBrains plugin.
It is a duplicate logic that can be replaced by plugin problem level remapping. 

I see the rationale behind this. 

However, it should be noted that any client code that uses the `structure` library must now explicitly establish plugin level remapping for JetBrains plugins.

novotnyr:
This might impact @hsz and [IntelliJ Platform Gradle Plugin Pl](https://github.com/JetBrains/intellij-platform-gradle-plugin).

novotnyr (approved):

YannCebron (approved):

PR #1164, opened 2024-09-25
Do not resolve Product Info layout components without actual files in the distribution

novotnyr:
Some distributions contain `product-info.json` with layout components that do not exist in the filesystem. 

For example, this is the layout entry:

```
    {
      "name": "org.jetbrains.plugins.emojipicker",
      "kind": "plugin",
      "classPath": [
        "plugins/emojipicker/lib/emojipicker.jar"
      ]
    }
```

However,  no `plugins/emojipicker/lib/emojipicker.jar` is available in the Windows distribution, as this plugin is Linux-only.

## Issues

- Platform-wide `ResourceResolver` is pointing to invalid files which will break resource resolution. Especially, when a plugin is XIncluding another descriptor that should be present in such resolver, resolution in a nonexistent JAR file will throw an exception.
- Parsing of such layout elements might lead to nonexistent classpath.

## Solution

- In platform-wide `ResourceResolver`, do not include JAR files that do not exist in the Platform distribution in the filesystem.
- When parsing a layout element, omit it when any of the declared classpaths (`classPath` elements) is not resolved in the Platform distribution in the filesystem.

## See also

See also [MP-6708](https://youtrack.jetbrains.com/issue/MP-6708)

### DIFF
--- a/intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/ProductInfoBasedIdeManager.kt
+++ b/intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/ProductInfoBasedIdeManager.kt
@@ -44,6 +44,8 @@ private val LOG: Logger = LoggerFactory.getLogger(ProductInfoBasedIdeManager::cl
 class ProductInfoBasedIdeManager : IdeManager() {
   private val productInfoParser = ProductInfoParser()
 
+  private val excludeMissingProductInfoLayoutComponents = true
+
   /**
    * Problem level remapping used for bundled plugins.
    */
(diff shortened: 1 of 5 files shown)


intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/ProductInfoBasedIdeManager.kt
```diff
@@ -155,16 +174,24 @@ class ProductInfoBasedIdeManager : IdeManager() {
...
+    }
+    return NamedResourceResolver(layoutComponent.name, CompositeResourceResolver(itemJarResolvers))
+  }
+
+  private fun logUnavailableClasspath(failedComponents: List<ResolvedLayoutComponent>) {
+    val logMsg = failedComponents.joinToString("\n") {
+      val cp = it.getClasspaths().joinToString(", ")
+      "Layout component '${it.name}' has some nonexistent 'classPath' elements: '$cp'"
     }
+    LOG.atDebug().log(logMsg)
```
YannCebron, line 194: why debug and not warn?

YannCebron: It could still also affect verification results in theory, right? And whilte it's not actionable, it's important for user/us to understand what might be possibly wrong with the given IDE distribution (and JB does not control all of them).

YannCebron (approved):

PR #1188, opened 2024-11-25
Consolidate resolution logic for missing layout component files

novotnyr:
In nightly versions of the Platform, the `product-info.json` contains layout components with `classpath` entries that do not exist in the filesystem. 

For example:
```
"layout": [
  {
    "name": "intellij.qodana.sarif",
    "kind": "moduleV2",
    "classPath": [
      "plugins/qodana/lib/modules/intellij.qodana.sarif.jar"
    ]
  }
]
```

- Consolidate logic resolution for both `ResourceResolver` and `Resolver` in the Structure library.
- By default, when a such missing classpath entries occur, skip the whole layout component. Then, log accordingly at _warn_ level.
- Add a CLI parameter switch `-missing-layout-classpath-file` to configure behavior from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,15 @@ Here is the full syntax of the command:
     A file that contains patterns of problems that will be reflected in report. All other problems will be ignored. Applied to short problem description.
     The file must contain lines in form: `<plugin_xml_id_regexp_pattern>:<plugin_version_regexp_pattern>:<problem_description_regexp_pattern>`
 
+* `-missing-layout-classpath-file`
+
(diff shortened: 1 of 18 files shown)


YannCebron (changes requested):
`MissingLayoutFileMode` should probably be configurable via cmdline switch

novotnyr:
> `MissingLayoutFileMode` should probably be configurable via cmdline switch

Added a CLI switch `-missing-layout-classpath-file`.

YannCebron (approved):

PR #1195, opened 2024-12-10
Mark module dependency as optional

LChernigovskaya:
- All `<content>` modules dependencies are now marked as _optional_.
- The modules from plugin `<dependencies>` tag are marked as required. In general, all plugin dependencies (specified in the `plugin.xml` are required - there is no difference if it's dependency on a plugin or on a module.


chrkv (approved):

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginV2Dependency.kt
```diff
@@ -4,12 +4,10 @@
 
 package com.jetbrains.plugin.structure.intellij.plugin
 
-data class PluginV2Dependency(private val id: String) : PluginDependency {
-  override fun getId() = id
+data class PluginV2Dependency(override val id: String, override var isOptional: Boolean = false) : PluginDependency {
+  override val isModule = id.startsWith(INTELLIJ_MODULE_PREFIX)
```
YannCebron, line 8: I'm not sure that is always the case

novotnyr: I have seen this issue and researched it. This is a pickup from previous implementation with unknown semantics.

novotnyr:
Refactored code to retain immutability of `PluginDependency` class hierarchy. Preserving semantics.

PR #1201, opened 2024-12-24
Reuse filesystem accessed by different uri to the same file

fp7:
When a jar file was opened with different uris pointing to the same file they would return the same file system because the jdk normalizes the uri, eg jar:file:/tmp/../tmp/file.jar becomes jar:file:/tmp/file.jar.

Therefore the number of users of one filesystem was incorrect and it might have been closed while still in use.

This occured eg. in mono repos with multiple intellij plugins and parallel builds.


fp7:
Sorry, I somehow completely missed the other tests. I fixed the failing ones and made sure that a `./gradlew build` works on my machine.

novotnyr:
This is a very nice catch and an improvement!

Thanks for your contribution.

PR #1206, opened 2025-01-19
Support inline <module> declarations with CDATA in plugin descriptor

novotnyr:
- Support v2 content module declarations in CDATA content.
- Support CDATA-based text representations in the v2 plugin module bean in the `content/module` elements.
- Until now, only file-based plugin descriptors were supported. Introduce a `DescriptorResource` indicating a plugin descriptor backed by an arbitrary `InputStream` identified by `URI` (usually a `file:` based).
- Introduce loading of modules from arbitrary `DescriptorResource` both in IDE plugins and `PkuginCreator`.
- Introduce two types of modules: file-based module (as per original spec) and an `InlineModule` backed by CDATA. Resolve both types of modules when creating a plugin.
- Introduce `PluginModuleResolver` that extracts away the v2 plugin module parsing.
- Introduce a _warning_ plugin problem `ModuleDescriptorProblem` indicating an issue with inline CDATA declaration.

See [MP-7092](https://youtrack.jetbrains.com/issue/MP-7092)

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginManager.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginManager.kt
@@ -449,19 +449,6 @@ class IdePluginManager private constructor(
     }.pluginCreationResult
   }
 
-  fun createInlineModule(
-    descriptorResource: DescriptorResource,
-    pluginFile: Path,
-    ideVersion: IdeVersion,
-    problemResolver: PluginCreationResultResolver = IntelliJPluginCreationResultResolver()
-  ): PluginCreationResult<IdePlugin> {
-    //FIXME parent plugin is null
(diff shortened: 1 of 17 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginManager.kt
```diff
@@ -504,12 +504,13 @@ class IdePluginManager private constructor(
 
   private fun getDescriptorResource(module: InlineModule, pluginFile: Path, descriptorPath: String): DescriptorResource {
     // FIXME descriptor path is not relative to the pluginFile JAR
```
YannCebron, line 506: what about FIXME?

novotnyr: Replaced by TODO and corresponding issue. The URI of the DescriptorResource is currently used only for debugging/logging scenarios. Making this a proper implementation will be out of scope for this PR. Created an additional issue (MP-7244) to handle this.

YannCebron (approved):

PR #1221, opened 2025-02-18
Support module aliases in class resolvers when used as dependencies

novotnyr:
A plugin might declare a dependency on a V1 plugin via V2 semantics.

```
<dependencies>
          <plugin id="com.intellij.platform.ide.provisioner" />
</dependencies>       
````

However, such plugin isan alias of the Core plugin. Add support for such scenario when resolving classes.

In `ProductInfoClassResolver`, group all `productModuleV2` from the Product Info under a single logical `Resolver`. In addition, maintain the list of mappings between product layout entries and their actual resolvers.

Thus, `com.intellij.platform.ide.provisioner` maps to the `Resolver` of the Core plugin with the nested resolvers still being available by their corresponding IDs.

### DIFF
--- a/intellij-plugin-structure/structure-ide-classes/src/main/java/com/jetbrains/plugin/structure/ide/classes/resolver/ProductInfoClassResolver.kt
+++ b/intellij-plugin-structure/structure-ide-classes/src/main/java/com/jetbrains/plugin/structure/ide/classes/resolver/ProductInfoClassResolver.kt
@@ -46,7 +46,7 @@ class ProductInfoClassResolver(
     }
   }
 
-  private fun List<ClasspathableResolver>.resolveCorePlugin(): Map<String, NamedResolver> {
+  private fun List<KindedResolver>.resolveCorePlugin(): Map<String, NamedResolver> {
     val corePluginResolver = find { it.name == CORE_IDE_PLUGIN_ID } ?: return emptyMap()
     val resolvers = mutableMapOf<String, NamedResolver>()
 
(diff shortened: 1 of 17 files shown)

### CODE REVIEW

File: intellij-plugin-structure/structure-ide-classes/src/main/java/com/jetbrains/plugin/structure/ide/classes/resolver/ProductInfoClassResolver.kt [hunk 1]
```diff
@@ -29,34 +30,80 @@ private val LOG: Logger = LoggerFactory.getLogger(ProductInfoClassResolver::clas
...
+      resolvers += resolveCorePlugin()
+      resolvers += resolveNonCoreAndNonProductModules()
+    }
+  }
+
+  private fun List<ClasspathableResolver>.resolveCorePlugin(): Map<String, NamedResolver> {
+    val corePluginResolver = find { it.name == CORE_IDE_PLUGIN_ID } ?: return emptyMap()
+    val resolvers = mutableMapOf<String, NamedResolver>()
+
+    val productModules = filter { it.kind == "productModuleV2" }
```
YannCebron, line 53: why use String constants here? see also below

novotnyr: Replaced with constants. Good catch!

YannCebron (approved):

PR #1224, opened 2025-02-27
Send warning if the description doesn't start with latin symbols

LChernigovskaya:
#MP-7198


novotnyr (changes requested):

LChernigovskaya:
Superseded by #1224

PR #1244, opened 2025-04-07
fix false-positive 'No dependencies' error and deprecate confusing 'IdePlugin::isV2'

chashnikov:
This pull request fixes [MP-7413](https://youtrack.jetbrains.com/issue/MP-7413) and [MP-7414](https://youtrack.jetbrains.com/issue/MP-7414),  and also deprecates 'IdePlugin::isV2' property which may be confusing.

### DIFF
--- a/intellij-feature-extractor/src/test/java/com/intellij/featureExtractor/MockIdePlugin.kt
+++ b/intellij-feature-extractor/src/test/java/com/intellij/featureExtractor/MockIdePlugin.kt
@@ -46,7 +46,9 @@ data class MockIdePlugin(
   override val useIdeClassLoader = false
   override val classpath: Classpath = Classpath.EMPTY
   override val isImplementationDetail = false
+  @Deprecated("See IdePlugin::isV2")
   override val isV2: Boolean = false
+  override val hasPackagePrefix: Boolean = false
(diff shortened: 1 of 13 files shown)


LChernigovskaya:
On the Marketplace side we use `isV2` to understand should we check if the plugin has any explicit dependencies or not, because if it's not v2 plugin and there are no dependencies we consider the plugin is compatible only with IDEA.
What should we check instead, `hasPackagePrefix`?

chashnikov:
> On the Marketplace side we use `isV2` to understand should we check if the plugin has any explicit dependencies or not, because if it's not v2 plugin and there are no dependencies we consider the plugin is compatible only with IDEA. What should we check instead, `hasPackagePrefix`?

The logic used in the platform is located [here](https://github.com/JetBrains/intellij-community/blob/b41a4084da5521effedd334e28896fd9d07410da/platform/core-impl/src/com/intellij/ide/plugins/PluginLoadingResult.kt#L158). I've updated code in `LegacyPluginAnalysis.isLegacyPlugin` in this commit to match this logic. We can move this class to intellij-structure library to reuse it in the Marketplace.

YannCebron (approved):

novotnyr (approved):

LChernigovskaya (approved):

PR #1259, opened 2025-04-30
Use Caffeine to cache JAR filesystem instances

novotnyr:


novotnyr:
Abandoned in favor of #1260.

PR #1270, opened 2025-05-02
Cache `DependencyNode` and `DeprecationInfo` instances

VladRassokhin:
Based on the profiling information:

- Cache `DependencyNode` instances when converting _Structure_ dependency graph to the Plugin Verifier dependency graph
- Reuse `DeprecationInfo` instances for more common cases


novotnyr:
@VladRassokhin thank you for your contribution!

PR #1280, opened 2025-05-19
Allow declaring `com.intellij.*` modules only for JetBrains plugins

novotnyr:
Only JetBrains plugins can declare v1 module (plugin alias) with specific prefixes:

- com.intellij
- org.jetbrains
- intellij

Changes:

- Introduce `ProhibitedModuleExposed` plugin problem. It is an _unacceptable warning_.
- For JetBrains plugins, remap this plugin problem to `ignore`, essentially suppressing it.

See [MP-7447](https://youtrack.jetbrains.com/issue/MP-7447)


intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/ExposedModulesTest.kt
```diff
@@ -0,0 +1,89 @@
...
+import com.jetbrains.pluginverifier.PluginVerificationResult
+import com.jetbrains.pluginverifier.options.PluginParsingConfiguration
+import com.jetbrains.pluginverifier.options.PluginParsingConfigurationResolution
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
+import org.junit.Test
+
+class ExposedModulesTest  : BasePluginTest() {
+  @Test
+  fun `plugin is a third-parts vendor and exposes JetBrains modules`() {
```
LChernigovskaya, line 19: `third-party`

LChernigovskaya (approved):

intellij-plugin-verifier/verifier-test/src/test/java/com/jetbrains/pluginverifier/tests/ExposedModulesTest.kt
novotnyr: Good spot!

PR #1315, opened 2025-07-12
Extract theme loading to a separate class

novotnyr:
Introduce `PluginThemeLoader` as a standalone class responsible for plugin theme loading.

Extract it away from `PluginCreator`.

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/loaders/PluginThemeLoader.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/loaders/PluginThemeLoader.kt
@@ -20,7 +20,7 @@ private const val INTELLIJ_THEME_EXTENSION = "com.intellij.themeProvider"
 class PluginThemeLoader {
   private val json = jacksonObjectMapper()
 
-  fun load(plugin: IdePlugin, document: Path, descriptorPath: String, resolver: ResourceResolver, problemRegistrar: ProblemRegistrar): Result {
+  fun load(plugin: IdePlugin, descriptorPath: Path, resolver: ResourceResolver, problemRegistrar: ProblemRegistrar): Result {
     val themePaths = plugin.extensions[INTELLIJ_THEME_EXTENSION]?.mapNotNull {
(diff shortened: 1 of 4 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/loaders/PluginThemeLoader.kt
```diff
@@ -0,0 +1,70 @@
...
+    for (themePath in themePaths) {
+      val absolutePath = if (themePath.startsWith("/")) themePath else "/$themePath"
+      when (val resolvedTheme = resolver.resolveResource(absolutePath, descriptorPath)) {
+        is ResourceResolver.Result.Found -> resolvedTheme.use {
+          val theme = resolvedTheme.use {
+            runCatching {
+              json.readValue(it.resourceStream, IdeTheme::class.java)
+            }.getOrElse {
+              problemRegistrar.unableToRead(descriptorPath, themePath)
+              return NotFound
```
harrytmthy, line 39: Just a thought on error-handling consistency and resilience:

The current behavior uses `return NotFound` on JSON parsing errors, but continues processing in the `NotFound` and `Failed` branches (as Copilot pointed out).

Rather than enforcing early returns across all cases, it might be worth considering the opposite: gracefully skipping individual theme failures and proceeding with the rest. That would mean changing the `.use {}` block to return `null` instead of exiting early, then adding the result to `themes` only if parsing succeeds:

```kotlin
val theme = resolvedTheme.use {
    runCatching {
        json.readValue(it.resourceStream, IdeTheme::class.java)
    }.getOrElse {
        problemRegistrar.unableToRead(descriptorPath, themePath)
        return@use null
    }
}
theme?.let { themes.add(it) }
```

This way, all error branches (missing file, corrupted JSON, resolution failure) behave consistently: they log the issue and continue. That could make the plugin loader more resilient to partial failures, especially if multiple themes are defined 🙂

novotnyr: Thanks for quick, thorough and thoughtful insights. That really helps.

PR #1319, opened 2025-07-17
Check the plugin name contains only allowed symbols

LChernigovskaya:
Introduce `InvalidPluginName` as an error for plugins that do not have a valid name.

This is an _error_ for new IntelliJ plugins and a _warning_ for existing IntelliJ plugins.

Note that plugin names that contain newlines will be automatically checked by the new rule, instead of `ContainsNewLine`.

See [MP-7202](https://youtrack.jetbrains.com/issue/MP-7202)


novotnyr:
Added some sidechecks and generators for valid XML strings. On top of that, LGTM.

kesarevs (approved):

PR #1347, opened 2025-09-30
intellij-structure: support 'visibility' and 'namespace' attributes when parsing plugin content modules (IJPL-206556)

chashnikov:


novotnyr (commented):
What about nullable namespaces? Isn't the namespace inherited from the parent, high up to the plugin ID as a fallback?

novotnyr (approved):

PR #1465, opened 2026-03-25
Fix false positive for Java 8 default method delegation in internal API check

marinaglancy:
When a Kotlin class implements a Java interface with a default method without overriding it, the Kotlin compiler generates a stub method that delegates via INVOKESPECIAL to the interface default. The verifier scans this generated method's return type and flags it as internal API usage, even though the plugin never references the internal type.

Extended isKotlinDefaultMethod() to detect this Java 8 default method delegation pattern, in addition to the existing Kotlin delegation detection.

Fixes https://youtrack.jetbrains.com/issue/MP-8037


marinaglancy:
I added a second commit because even if the method is overridden, the verification was still failing. Now it skips but it may be too aggressive (see more details in the issue)

novotnyr (approved):
Thanks, looking good!

fgaraba:
As per the discussion on the ticket, I will close this PR. Hopefully your more specific issue with uploading your plugin is resolved!

PR #1476, opened 2026-04-15
Reduce plugin-structure allocation and FS overhead in plugin loading paths

hsz:
## Summary

This PR trims plugin-structure changes down to the small set that still has a clear performance signal in the `buildPlugin --rerun-tasks` profile.

The focus is on reducing avoidable filesystem work and allocation churn in hot plugin-loading paths:
- `jar:file:` URI creation
- plugin descriptor creation
- content-module scanning
- recursive XInclude resolution

### Avoid extra filesystem work for JAR URIs
In `Jars.kt`:
- stop resolving JAR paths through `toRealPath()`
- normalize the absolute local path once and build the `jar:file:` URI from that

This removes an unnecessary filesystem hit on the common local-JAR path.

### Reuse stateless helpers in `PluginCreator`
In `PluginCreator.kt`:
- reuse `PluginDescriptorParser`
- reuse `PluginBeanValidator`
- reuse `PluginBeanToIdePluginConverter`
- reuse plugin verifier instances

Previously these were created repeatedly during plugin parsing. They are stateless and safe to reuse.

### Skip content-module scanning when it is clearly unnecessary
In `LibDirectoryPluginLoader.kt`:
- if the plugin declares no content modules, skip `ContentModuleScanner` entirely
- use `Classpath.EMPTY` for that branch

This avoids opening/scanning additional plugin JARs when the descriptor already tells us there is nothing to resolve there.

### Reduce allocation churn in XInclude resolution
In `XIncluder.kt`:
- replace `Stack` with a lightweight list-backed recursion stack
- avoid rebuilding trivial leaf elements
- reduce list churn when wrapping included content
- simplify detach/copy handling for selected content

This keeps behavior unchanged while making the recursive include path cheaper.

## Why

Recent Windows JFR profiles of the plugin template `buildPlugin` flow showed PV-owned work in:
- `PluginCreator`
- `PluginBeanToIdePluginConverter`
- `ContentModuleScanner`
- `XIncluder`

The broader earlier experiments did not produce a stable win, so this PR keeps only the subset with a defensible signal.

## Profiling result

Compared on Windows using the same 60s JFR recording setup for `buildPlugin --rerun-tasks`:

- `GCPhaseParallel`: `79248 -> 26523`
- `ObjectAllocationSample`: `6537 -> 4287`
- `PromoteObjectInNewPLAB`: `7535 -> 4494`
- `ExecutionSample`: `302 -> 133`
- `JavaMonitorWait`: `112 -> 10`

Sampled stack presence also dropped:
- PV hits: `48 -> 10`
- `XIncluder` disappeared from sampled stacks
- `JarsKt` disappeared from sampled stacks

The remaining dominant cost is mostly Gradle + Windows filesystem startup work, not plugin-structure.

## Testing

Ran:
- `./gradlew :tests:test --tests com.jetbrains.plugin.structure.jar.JarsTest --tests com.jetbrains.plugin.structure.xinclude.XIncluderTest --no-daemon --console=plain`
- `./gradlew :structure-intellij:test --tests com.jetbrains.plugin.structure.intellij.plugin.loaders.JarOrDirectoryPluginLoaderTest --tests com.jetbrains.plugin.structure.intellij.plugin.PluginCreatorTest --no-daemon --console=plain`

Both passed.


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/xinclude/XIncluder.kt
```diff
@@ -47,13 +48,12 @@ class XIncluder private constructor(private val resourceResolver: ResourceResolv
     if (isIncludeElement(document.rootElement)) {
       throw XIncluderException(listOf(startEntry), "Invalid root element ${document.rootElement.getElementNameAndAttributes()}")
     }
-    val bases = Stack<XIncludeEntry>()
-    bases.push(startEntry)
+    val bases = ArrayList<XIncludeEntry>(4).apply { add(startEntry) }
```
novotnyr, line 51: Why `4`?

novotnyr (approved):

novotnyr:
This looks very promising! Thanks for the PR.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/PluginCreator.kt
```diff
@@ -36,6 +36,15 @@ internal class PluginCreator private constructor(
...
     private val themeLoader = PluginThemeLoader()
+    private val descriptorParser = PluginDescriptorParser()
+    private val beanValidator = PluginBeanValidator()
+    private val beanToPluginConverter = PluginBeanToIdePluginConverter()
+    private val legacyIntelliJIdeaPluginVerifier = LegacyIntelliJIdeaPluginVerifier()
+    private val projectAndApplicationListenerAvailabilityVerifier = ProjectAndApplicationListenerAvailabilityVerifier()
+    private val serviceExtensionPointPreloadVerifier = ServiceExtensionPointPreloadVerifier()
+    private val statusBarWidgetFactoryExtensionPointVerifier = StatusBarWidgetFactoryExtensionPointVerifier()
+    private val k2IdeModeCompatibilityVerifier = K2IdeModeCompatibilityVerifier()
+    private val exposedModulesVerifier = ExposedModulesVerifier()
```
fgaraba, line 47: Are you sure this is a semantically equivalent refactor? If we have state in any of these, then we can end up with some hard-to-debug bugs.

And the proposed benefit is probably negligible (just creating a couple of objects is almost free).

On a quick glance it does seem that these are all stateless, though, so feel free to have it as is.

fgaraba (approved):

PR #1486, opened 2026-04-30
Migrate JAXB to Jakarta namespace

kesarevs:
We need to migrate from Java EE to Jakarta in Marketplace because newer versions of Hibernate brings a new `org.glassfish.jaxb:jaxb-runtime` that is not backward compatible.

I suggest not rushing this and first evaluating how difficult it will be to migrate other clients to a newer runtime. Please add more people to review. 

I also recommend releasing the important upcoming changes first, and then handling the migration in a separate version to avoid getting stuck, since the full migration in Marketplace will take some time. Maybe we can release it as dev build?

To check how hard the migration might be the following commands might be usefull:
```
./gradlew dependencyInsight --dependency org.glassfish.jaxb:jaxb-runtime --configuration runtimeClasspath
```

### DIFF
--- a/intellij-plugin-structure/gradle/libs.versions.toml
+++ b/intellij-plugin-structure/gradle/libs.versions.toml
@@ -5,7 +5,7 @@ asm = "9.9.1"
 commons-compress = "1.28.0"
 evo-inflector = "1.3"
 jackson-dataformat-yaml = "2.14.2"
-jaxb = "2.3.1"
+jaxb = "4.0.2"
 jdom = "2.0.6.1"
 jimfs = "1.3.1"
 kotlinx-metadata = "0.9.0"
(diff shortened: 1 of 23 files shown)


kesarevs:
Hi, folks! Could you please estimate when we can release these, so I could plan other parts of these migrations?

fgaraba (approved):
I see nothing problematic about the PR from a plugin verifier perspective, but I am not too familiar with the other services that integrate it, where this may cause some problems.

novotnyr (approved):

novotnyr:
Why not use [4.0.5](https://github.com/jakartaee/jaxb-api/releases/tag/4.0.5)?

kesarevs:
> Why not use [4.0.5](https://github.com/jakartaee/jaxb-api/releases/tag/4.0.5)?

It's just aligned with other components. Let's update it later after the release to simplify the migration

PR #1487, opened 2026-05-10
Fix ClosedFileSystemException in parallel verification (MP-7267)

sh41:
Both the description below and the fix came from Claude. I haven't been able to build the jar locally to test, but that's due to my environment rather than the code itself. 

==== AIA GENERATED BELOW THIS LINE ====

When running multiple plugin verifications in parallel (e.g. 4 IDEs simultaneously), the verifier intermittently crashes with `ClosedFileSystemException` during bundled plugin scanning. This is tracked as [MP-7267](https://youtrack.jetbrains.com/issue/MP-7267).

The crash originates in `FsHandlerFileSystemProvider.newDirectoryStream()`, which delegates to the underlying `ZipFileSystemProvider` but returns the raw `ZipPath` entries **without wrapping them in `FsHandlerPath`**. These unwrapped paths bypass `FsHandleFileSystem`'s reopen-on-close mechanism, so when the `SingletonCachingJarFileSystemProvider`'s Caffeine LRU cache evicts the entry (due to concurrent pressure from other verifier threads), the underlying `ZipFileSystem` is closed while iteration is still in progress.

The failure trace:

```
Files.list(root)           ← returns raw ZipPath entries
  → filter { hasExtension("xml") }
    → Files.isRegularFile(rawZipPath)
      → ZipFileSystem.ensureOpen()
        → ClosedFileSystemException
```

Every other method in `FsHandlerFileSystemProvider` correctly uses `path.unwrapped` for input and handles results safely through the `FsHandleFileSystem` reopen layer. `newDirectoryStream` is the only method that leaks raw delegate paths back to callers.

### Fix

Wrap the `DirectoryStream` returned by the delegate so that each entry from the iterator is wrapped in `FsHandlerPath(fs, delegateEntry)`. This ensures all returned paths route back through `FsHandlerFileSystemProvider`, which checks `delegatePath.fileSystem.isClosed` and transparently reopens the filesystem if needed — the same safety net every other method already provides.

### Testing

This race condition requires concurrent cache pressure to reproduce: 4+ parallel verifier threads, 256+ unique JAR paths in flight, and unlucky eviction timing. It was observed in CI with 4 simultaneous IDE verifications. The fix is structurally correct — it closes the only gap where raw `ZipPath` entries could escape the `FsHandleFileSystem` wrapper.

---
# Plugin Verifier CI Failure: `ClosedFileSystemException` in Parallel Verification

**Date**: 2026-05-10  
[...]

PR #1488, opened 2026-05-11
Fix closed filesystem exception

fgaraba:
Attempts to fix https://youtrack.jetbrains.com/issue/MP-7468/Plugin-Verifier-ClosedFileSystemException-when-verifying-large-number-of-plugins

This is an exception due to a race condition where an underlying resource is closed on one thread, so reproduction is non-trivial. This includes two related changes, one from https://github.com/JetBrains/intellij-plugin-verifier/pull/1487 @sh41 and one from me.


fgaraba:
Temporarily removed @novotnyr 's commits to check what change solves the race condition (it's hard to reproduce locally).

PR #1489, opened 2026-05-11
Do not report OverrideOnly violations on static and companion calls

abrooksv:
Implement https://youtrack.jetbrains.com/issue/MP-7233/ by checking if the target method is not static and not a companion object

Also disabled the Javadoc tasks on the verifier test projects because it was preventing Gradle from running the tests through the IDE


novotnyr (approved):
Thank you for spotting this corner-case!

novotnyr:
`NonOverridableMethodUsageFilter` is using class resolver as an instance variable.

```kotlin
private val ktClassResolver = KtClassResolver()
```

Just FTR, this class serves as a cache. If there are multiple instances, such cache would be duplicated. However, the second such usage - in Kotlin Internal Usage filter is disabled now.

fgaraba:
From what I see there is a bit of an asymmetry in how it handles java vs kotlin classes.

`isStatic` works in both languages' case, but `isKotlinObject`  can be handled with a simple check for java final classes (that kotlin objects compile to, and they also can't be subclasses). So we can make it a bit more generic. 

I have added this change and a test for that, but unfortunately can't push to your fork. I guess I will just merge your PR and create a follow up. 

Thanks for the code!

fgaraba (approved):

PR #1502, opened 2026-05-27
Propagate companion annotations to @JvmStatic methods and fields (IJPL-245371)

knisht:
The following `foo` method must be marked as `@Internal` as well.

```kotlin
interface A {
  @Internal
  companion object {
    @JvmStatic fun foo() {}
  }
}
```

This applies for any annotation on the `companion object`. For example, `@Experimental` as well.

See [IJPL-245371](https://youtrack.jetbrains.com/issue/IJPL-245371).


fgaraba (approved):

novotnyr (approved):

novotnyr:
Thank you for spotting this!

PR #1506, opened 2026-06-03
Merge semantics for OverrideOnly in Java final class and Kotlin object

fgaraba:
Follow up of #1489 

isStatic works in both languages' case, but isKotlinObject can be handled with a simple check for java final classes (that kotlin objects compile to, and they also can't be subclasses). So we can make it a bit more generic.


novotnyr:
Looks nice and is even more simple!

novotnyr (approved):

PR #1514, opened 2026-06-15
Close ZipFile eagerly in ZipFileHandler instead of relying on GC to r…

fgaraba:
@hsz  has found a leaking file handler, thank you for your contribution!

This is a follow up to https://github.com/JetBrains/intellij-plugin-verifier/pull/1508


novotnyr:
I need additional explanation why is this happening. This is essentially a reversal of optimisations in January '26 in a74dc274764dd719ed31e00070715e631877d5c4 by @VladRassokhin that speeds up the validation.

Why is this being reverted?

hsz:
The optimization is useful, but it hides file-handle ownership.
`ZipFileHandler` keeps a `ZipFile` open after `containsEntry` / `handleEntry` returns, and relies on weak/soft references plus `Cleaner` to close it later. That close is nondeterministic. On Windows, an open `ZipFile` locks the `.zip` / `.jar`, so Gradle cleanup can fail even though the code path looks finished.

novotnyr (approved):

PR #1515, opened 2026-06-16
Improve Filesystem handle delegate race corner-cases

novotnyr:
Make delegate accounting track the actual delegate object, not just the JAR key, and notify the cache provider when `FsHandleFileSystem` swaps delegates during reopen.

The problem is this sequence:

1. `fsOne` and `fsTwo` both wrap the same delegate `D`; cache ref count is 2.
2. `D` is closed underneath them.
3. `fsOne` is used again, so it reopens and now points to `D2`.
4. The cache-level ref count still says 2, but it no longer knows that `fsOne` holds `D2` and `fsTwo` still holds stale `D`.
5. If `fsOne` is evicted first, it decrements the count to 1 and does not close `D2`.
6. If `fsTwo` is evicted second, it decrements to 0 against stale `D`, which is already closed, so `D2` can be left open.

The issue is that one wrapper migrates to a reopened delegate while another wrapper still references the old closed delegate.

### DIFF
--- a/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/jar/CachingJarFileSystemProvider.kt
+++ b/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/jar/CachingJarFileSystemProvider.kt
@@ -116,26 +116,54 @@ class CachingJarFileSystemProvider(
     val jarUri = jarPath.toJarFileUri()
     val key = jarUri.toString()
 
-    synchronized(key.intern()) {
-      var fs = fsCache.getIfPresent(key)
-      if (fs != null) {
-        if (fs.increment(expectedClients)) {
-          logReusedFs(key)
-        } else {
+    // Retry because increment() must run outside the per-key lock: while it is running,
+    // Caffeine may evict the cached handle or another thread may replace it. In that case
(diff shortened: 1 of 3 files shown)


intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/jar/CachingJarFileSystemProvider.kt
```diff
@@ -89,28 +116,54 @@ class CachingJarFileSystemProvider(
...
-      if (fs != null) {
-        if (fs.increment(expectedClients)) {
-          logReusedFs(key)
-        } else {
-          val jarFs = delegateJarFileSystemProvider.getFileSystem(jarPath).also {
-            LOG.debug("Recreating an already closed a filesystem handler for <{}> (Cache size: {})", key, fsCache.estimatedSize())
+    // Retry because increment() must run outside the per-key lock: while it is running,
+    // Caffeine may evict the cached handle or another thread may replace it. In that case
+    // undo the acquired references, re-read the cache, and try again against the current entry.
+    while (true) {
```
fgaraba: Maybe add a bound here with an error log, to prevent a hypothetical live lock ever causing hard to debug issues in the future.

fgaraba (commented):
Besides the bound on the while true loop, it looks good!

PR #1527, opened 2026-06-24
Bump libs and cleanup exposed dependency

kesarevs:


kesarevs:
It's getting wild, I'll split the changes.

PR #1545, opened 2026-07-06
Reduce object allocation during check-trunk

fgaraba:
API compatibility checks have started timing out against latest trunk IDE, so I have taken a deeper look at the JFR recordings.

Unfortunately most of the issues we have experiences recently all come down to an insanely huge allocation rate with a tendency for these objects to remain alive.

I have taken a look at the most allocation-heavy locations and made several, low-hanging fruit fixes (with the help of some LLM), that I deemed have a low risk of breaking anything.

### DIFF
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/ClassFileAsm.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/ClassFileAsm.kt
@@ -13,13 +13,14 @@ import org.objectweb.asm.tree.ClassNode
 import org.objectweb.asm.tree.InnerClassNode
 
 class ClassFileAsm(val asmNode: ClassNode, override val classFileOrigin: FileOrigin) : ClassFile {
-  override val location
-    get() = ClassLocation(
+  override val location: ClassLocation by lazy(LazyThreadSafetyMode.NONE) {
+    ClassLocation(
       name,
       signature?.takeIf { it.isNotEmpty() },
       Modifiers(asmNode.access),
       classFileOrigin
     )
+  }
 
   override val containingClassFile
     get() = this
@@ -33,8 +34,11 @@ class ClassFileAsm(val asmNode: ClassNode, override val classFileOrigin: FileOri
   override val javaPackageName
     get() = packageName.replace('/', '.')
 
-  override val methods
-    get() = asmNode.methods.asSequence().map { MethodAsm(this, it) }
+  private val _methods: List<MethodAsm> by lazy(LazyThreadSafetyMode.NONE) {
+    asmNode.methods.map { MethodAsm(this, it) }
+  }
+  override val methods: Sequence<Method>
+    get() = _methods.asSequence()
 
   override val fields
     get() = asmNode.fields.asSequence().map { FieldAsm(this, it) }
--- a/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodAsm.kt
+++ b/intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodAsm.kt
@@ -16,8 +16,8 @@ import org.objectweb.asm.tree.LocalVariableNode
 import org.objectweb.asm.tree.MethodNode
 
 class MethodAsm(override val containingClassFile: ClassFile, val asmNode: MethodNode) : Method {
-  override val location
-    get() = MethodLocation(
+  override val location: MethodLocation by lazy(LazyThreadSafetyMode.NONE) {
+    MethodLocation(
       containingClassFile.location,
       name,
       descriptor,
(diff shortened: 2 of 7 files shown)


VladRassokhin (approved):

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/ClassFileAsm.kt
```diff
@@ -13,13 +13,14 @@ import org.objectweb.asm.tree.ClassNode
 import org.objectweb.asm.tree.InnerClassNode
 
 class ClassFileAsm(val asmNode: ClassNode, override val classFileOrigin: FileOrigin) : ClassFile {
-  override val location
-    get() = ClassLocation(
+  private var _location: ClassLocation? = null
```
novotnyr: by lazy {}?

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/ClassFileAsm.kt
```diff
@@ -33,8 +34,9 @@ class ClassFileAsm(val asmNode: ClassNode, override val classFileOrigin: FileOri
   override val javaPackageName
     get() = packageName.replace('/', '.')
 
-  override val methods
-    get() = asmNode.methods.asSequence().map { MethodAsm(this, it) }
+  private var _methods: List<MethodAsm>? = null
```
novotnyr: by lazy {}?

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/MethodAsm.kt
```diff
@@ -16,15 +16,16 @@ import org.objectweb.asm.tree.LocalVariableNode
 import org.objectweb.asm.tree.MethodNode
 
 class MethodAsm(override val containingClassFile: ClassFile, val asmNode: MethodNode) : Method {
-  override val location
-    get() = MethodLocation(
+  private var _location: MethodLocation? = null
+  override val location: MethodLocation
+    get() = _location ?: MethodLocation(
```
novotnyr: by lazy{}?

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/annotation/AnnotationResolver.kt
```diff
@@ -26,6 +29,16 @@ private val LOG: Logger = LoggerFactory.getLogger(AnnotationResolver::class.java
...
 class AnnotationResolver(val annotation: FullyQualifiedClassName) {
 
+  /**
+   * Caches, per class [Resolver], whether a package's `package-info` class exists and is
+   * directly annotated with [annotation]. Most packages have no `package-info`, and resolving
+   * that miss through the composite resolver chain is expensive; the answer cannot change
+   * within a run because resolvers are immutable snapshots. Keys are weak: `AnnotationResolver`
+   * instances are long-lived singletons, while resolvers live only for a single verification.
+   */
+  private val packageInfoAnnotationCache: MutableMap<Resolver, ConcurrentHashMap<String, Boolean>> =
```
novotnyr: `Resolver` classes have no `hashCode()`/`equals()` semantics defined. If you are comparing with `==` identity, I suggest Guava. 

```
Cache<Key, Value> cache = Caffeine.newBuilder()
.weakKeys()
.build();
```

novotnyr (changes requested):
Please rename the Pull Request so it can become a human-readable line in a changelog.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/resolution/ClassFileAsm.kt
fgaraba: `by lazy` synchronizes over the instance, but it is only accessed from a single thread. There is an option for thread safety `None` with an argument, but the other similar site had a very long and convoluted lambda, and at that point it felt overcomplicated a bit.

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/usages/annotation/AnnotationResolver.kt
fgaraba: `WeakHashMap` should also be identity-based, see the java doc: "This class is intended primarily for use with key objects whose equals methods test for object identity using the == operator."


Do you know of a functional difference between this and the caffeine version?

novotnyr: Differences:
- `WeakHashMap` switches to `equals()`, if available. However, this does not apply to `Resolver` subclasses and realistically never will.

We use consistently `Caffeine` as a semantic indicator of cache across the library, so I suggest to be consistent to indicate intent.

chrkv (approved):

novotnyr (approved):

PR #1546, opened 2026-07-07
Deduplication in ResolvedDependenciesGraph 

VladRassokhin:
Add global cache for all plugins, deduplicate edges (between different graphs)

### DIFF
--- a/intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/PluginVerifier.kt
+++ b/intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/PluginVerifier.kt
@@ -18,8 +18,8 @@ import com.jetbrains.pluginverifier.analysis.ExtractedJsonPluginAnalyzer
 import com.jetbrains.pluginverifier.analysis.ReachabilityGraph
 import com.jetbrains.pluginverifier.analysis.buildClassReachabilityGraph
 import com.jetbrains.pluginverifier.dependencies.DependenciesGraph
-import com.jetbrains.pluginverifier.dependencies.toResolved
 import com.jetbrains.pluginverifier.dependencies.ModuleVisibilityChecker
+import com.jetbrains.pluginverifier.dependencies.toResolved
 import com.jetbrains.pluginverifier.dymamic.DynamicPlugins
 import com.jetbrains.pluginverifier.filtering.ApiUsageFilter
 import com.jetbrains.pluginverifier.filtering.ExternalBuildClassesSelector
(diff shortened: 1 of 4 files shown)


intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/MultiplePluginVerifiersCache.kt
```diff
@@ -0,0 +1,18 @@
...
+
+package com.jetbrains.pluginverifier
+
+import java.util.concurrent.ConcurrentHashMap
+import java.util.function.Function
+
+/**
+ * Context for [com.jetbrains.pluginverifier.runSeveralVerifiers]
+ */
+class MultiplePluginVerifiersCache {
```
VladRassokhin: Name is discussable

fgaraba (approved):
Nice catch, did you go from what's the most common object in the live set?

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/MultiplePluginVerifiersCache.kt
novotnyr: PluginVerifierBatchCache? 

Or even, go further `PluginVerifierBatchContext`.

novotnyr (approved):
LGTM aside from the naming discussion.

intellij-plugin-verifier/verifier-intellij/src/main/java/com/jetbrains/pluginverifier/MultiplePluginVerifiersCache.kt
VladRassokhin: `PluginVerifierBatchContext` sounds better, indeed

VladRassokhin:
>most common object in the live set

I don't know since references are not counted.  (Probably could be somehow calculated from dump?)

I've dumped memory when verifier was in the process of generating HTML, so all `Verified` objects were present in memory and strong referencable.

I see that the number of `ResolvedDependencyEdge` reduced from 80.4M to just 52K after I've added `dedup` on them (was a fixup commit). Shallow size reduced from 1.9 GiB to 1.2 MiB.

Now all 14k 'ResolvedDependenciesGraph's have retained size of 4.8 GiB instead of 6.7GiB before last `dedup`. 
When we had OOMs, we had 80M `ResolvedDependencyEdge`, 8.6M `ResolvedDependencyNode` and 9.5M `ResolvedPluginDependency`. Now we have 52K edges, 9.3K nodes and 1.6K dependencies.

See two attached screenshots from YourKit.

memory after changes
memory in oom

VladRassokhin:
I've pushed yet another one `dedup` directly to master after checking "duplicate objects" YourKit inspection in the latest snapshot.

PR #1549, opened 2026-07-14
Reapply "Close ZipFile eagerly" to fix Windows file leak

fgaraba:
There was a very significant performance optimization that replaced deterministic ZIP closing with registering Cleaners that do it based on GC heuristics, making ZIP file reuse possible.

Unfortunately on windows it introduced some problems, but so did reverting it - the latter caused CI failures, so we decided to temporarily rather have the windows issues.

Now we have managed to get the CI to a more stable and performant state (again, thanks to @VladRassokhin ), so we can live with the increased runtime that reapplying this change causes -- the long CI jobs are still faster now than ever before.

### DIFF
--- a/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/zip/ZipFileHandler.kt
+++ b/intellij-plugin-structure/structure-base/src/main/kotlin/com/jetbrains/plugin/structure/base/zip/ZipFileHandler.kt
@@ -6,10 +6,6 @@ package com.jetbrains.plugin.structure.base.zip
 
 import java.io.File
 import java.io.IOException
-import java.lang.ref.Cleaner
-import java.lang.ref.Reference
-import java.lang.ref.SoftReference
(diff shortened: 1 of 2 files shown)


VladRassokhin:
How significantly does this rollback affect builds with thousands of plugins? AKA the build which compares compatibility of nightly and last stable IDE.

Also, separate system parameter could control whether ZIPs are closed earlier or not.

fgaraba:
It increases the runtime from ~45 mins to around ~90 mins. For reference, before that most of the runs were 2 and a half hours or more.

I can look into finding another approach, but this was discussed with @chrkv  and @novotnyr and they agreed that correctness should come first, and @hsz have notified us of the problem a long time ago.

VladRassokhin:
We discussed that and agreed to merge this MR for improved correctness on Windows and later introduce an option to have a deferred `ZipFile` closing mechanism, maybe only for Linux/macOS.

novotnyr (approved):

PR #1555, opened 2026-07-21
Bump Kotlin 2.2

fgaraba:
## Summary

- Bumps the Kotlin compiler from 2.0.21 to 2.4.10 across all five sub-builds. This unblocks the use of `com.jetbrains.intellij.platform:plugin-system-parser-impl` (compiled with Kotlin 2.4 metadata), which is being added to the `structure-intellij` version catalog as groundwork for a future parser migration, as well as some other issues ([MP-7810](https://youtrack.jetbrains.com/issue/MP-7810) Correctly handle 403 error from Marketplace in verifier service)
- Replaces the bytecode-shape heuristic in `KotlinMethods.isKotlinDefaultMethod()` with a semantically correct check against `@kotlin.Metadata` (via `kotlinx-metadata-jvm`). The old approach pattern-matched a specific `INVOKESTATIC → $DefaultImpls` instruction sequence that the Kotlin 2.x compiler no longer emits when `-Xjvm-default=all-compatibility` is active — causing the `check that all internal API violating usages are found` test to report four false-positive `NoInternalTypeUsage` findings after the compiler bump. The new approach asks whether the method appears in the class's own declared functions list in Kotlin metadata; if it doesn't, the compiler synthesized it - regardless of which lowering strategy produced the bytecode.
- Pins `apiVersion`/`languageVersion` in the five verifier-test fixture modules (`mock-plugin`, `before-idea`, `after-idea`, `additional-before-idea`, `additional-after-idea`) independently from the project's own language level, so the fixture bytecode shape doesn't drift with future toolchain bumps.

## Mechanical fixes required by the version bump

- `ide-diff-builder`: updated `kotlinx-serialization-runtime:1.0-M1-1.4.0-rc` (incompatible with the 2.4.x plugin) to `kotlinx-serialization-json:1.11.0`; removed dead `import kotlinx.serialization.stringify`.
- `plugins-verifier-service ValueType<T>`: added `: Any` upper bound — Kotlin 2.2 promotes to an error the case where a Kotlin class with a nullable `T` overrides a Java generic member that expects non-null `T`.
- `KtClassResolver`: replaced `cache.get(key) { nullable }` (rejected under Kotlin 2.2's stricter Caffeine overload resolution) with `getIfPresent` + conditional `put`.
- `CachingPluginDependencyResolverProvider`: replaced `cache.put(this.pluginId, r)` (`String?`) with `cache.put(this.id, r)` (`String`).
- Various `toLowerCase()`/`toUpperCase()` calls replaced with `lowercase()`/`uppercase()` (three files).

### DIFF
--- a/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
+++ b/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
@@ -21,9 +21,8 @@ class KtClassResolver {
 
   operator fun get(classNode: ClassNode): KtClassNode? {
     val signature: Signature = classNode.signature ?: return classNode.ktClassNode
-    return cache.get(signature) {
-      classNode.ktClassNode
-    }
+    return cache.getIfPresent(signature)
+      ?: classNode.ktClassNode?.also { cache.put(signature, it) }
   }
 
   private val ClassNode.ktClassNode: KtClassNode?
--- a/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
+++ b/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
@@ -17,12 +17,11 @@ private val LOG: Logger = LoggerFactory.getLogger(KtClassResolver::class.java)
 class KtClassResolver {
   private val cache = Caffeine.newBuilder()
     .maximumSize(16_384)
-    .build<Signature, KtClassNode>()
+    .build<Signature, KtClassNode?>()
 
   operator fun get(classNode: ClassNode): KtClassNode? {
     val signature: Signature = classNode.signature ?: return classNode.ktClassNode
-    return cache.getIfPresent(signature)
-      ?: classNode.ktClassNode?.also { cache.put(signature, it) }
+    return cache.get(signature) { classNode.ktClassNode }
   }
 
   private val ClassNode.ktClassNode: KtClassNode?
--- a/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
+++ b/intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
@@ -17,11 +17,23 @@ private val LOG: Logger = LoggerFactory.getLogger(KtClassResolver::class.java)
 class KtClassResolver {
   private val cache = Caffeine.newBuilder()
     .maximumSize(16_384)
-    .build<Signature, KtClassNode?>()
+    .build<Signature, KtClassNode>()
 
+  /**
+   * Deliberately *not* `cache.get(signature) { ... }`, even though that form is more concise.
+   *
+   * * a mapper returning `null` doesn't type-check against a non-null value type, and Caffeine
+   *   records nothing for a `null` result, so a class without Kotlin metadata is never cached and
+   *   every lookup for it takes the mapper path;
+   * * `Cache.get(key, mapper)` computes under a `ConcurrentHashMap` bin lock and runs the mapper
+   *   while holding it, whereas `getIfPresent` is lock-free. Verification runs this per class on
+   *   16 threads, so the always-missing keys above would serialize those threads on metadata
+   *   parsing. Measured at 16 threads on one hot key, the mapper form was ~7x slower.
+   */
   operator fun get(classNode: ClassNode): KtClassNode? {
     val signature: Signature = classNode.signature ?: return classNode.ktClassNode
-    return cache.get(signature) { classNode.ktClassNode }
+    return cache.getIfPresent(signature)
+      ?: classNode.ktClassNode?.also { cache.put(signature, it) }
   }
 
   private val ClassNode.ktClassNode: KtClassNode?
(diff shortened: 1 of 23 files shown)


intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -8,9 +8,10 @@ import com.github.benmanes.caffeine.cache.Cache
 import com.github.benmanes.caffeine.cache.Caffeine
 import com.jetbrains.pluginverifier.results.location.MethodLocation
 import com.jetbrains.pluginverifier.verifiers.resolution.Method
-import org.objectweb.asm.Opcodes
-import org.objectweb.asm.tree.AbstractInsnNode
-import org.objectweb.asm.tree.MethodInsnNode
+import kotlinx.metadata.jvm.KotlinClassMetadata
+import kotlinx.metadata.jvm.Metadata
+import kotlinx.metadata.jvm.signature
+import org.objectweb.asm.tree.AnnotationNode
```
fgaraba, line 20: This file is the only "non-mechanic" change in this PR, this should warrant a more thorough review.

It's necessary in order to make the test case pass (we test classes that are compiled by the project's kotlin compiler, so bumping the compiler version changes the test subject itself).

This may be a bit close coupling though, if you say so I can split it into a separate PR.

VladRassokhin (approved):

novotnyr:
There is no need to address `plugins-verifier-service`. This subproject is essentially archived. I'd revert the changes in that part.

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
```diff
@@ -21,9 +21,8 @@ class KtClassResolver {
 
   operator fun get(classNode: ClassNode): KtClassNode? {
     val signature: Signature = classNode.signature ?: return classNode.ktClassNode
-    return cache.get(signature) {
-      classNode.ktClassNode
-    }
+    return cache.getIfPresent(signature)
+      ?: classNode.ktClassNode?.also { cache.put(signature, it) }
```
novotnyr, line 36: Why this change? .get() with lambda is handling this semantic automatically.

intellij-plugin-structure/structure-ide-classes/src/main/java/com/jetbrains/plugin/structure/ide/classes/resolver/CachingPluginDependencyResolverProvider.kt
```diff
@@ -91,7 +91,7 @@ class CachingPluginDependencyResolverProvider(
   private fun Dependency.createResolverTree(): NamedResolver {
     return plugin?.createResolverTree()
       ?.let { (r, resolversToCache) ->
-        cache.put(this.pluginId, r)
+        cache.put(this.id, r)
```
novotnyr: Why this change? It is essentially about nullability? But the semantics might change

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
novotnyr: I have added unit test that covers this in all known combinations of Kotlin compiler behavior. The tests are ASM-driven, targeted for the detections, so they are Gradle Kotlin compiler independent.

intellij-plugin-verifier/verifier-core/src/main/java/com/jetbrains/pluginverifier/verifiers/method/KotlinMethods.kt
```diff
@@ -22,134 +23,89 @@ object KotlinMethods {
...
-        i++
-        continue
-      }
-      if (instruction.isKotlinIntrinsic()) {
-        i += 3 // Skip ALOAD, LDC, and INVOKESTATIC
-        continue
-      }
-      candidateOpcodes += instruction
-      i++
+    val isDeclaredByThisClass = kmClass.functions.any { function ->
```
novotnyr: replace with mapNotNull() and any()

novotnyr (changes requested):

novotnyr:
What is the relationship of this PR to #1394?

fgaraba:
@novotnyr I have rebased that old PR's branch to an up-to-date master, and fixed the blocker issue that prevented us from finishing it the last time.

So it's more of a "spiritual successor" -- after this is merged, the old one can be closed.

intellij-plugin-structure/structure-classes/src/main/java/com/jetbrains/plugin/structure/classes/utils/KtClassResolver.kt
fgaraba: `.get {}` with lambda no longer compiles, but I have fixed the unnecesary split

intellij-plugin-structure/structure-ide-classes/src/main/java/com/jetbrains/plugin/structure/ide/classes/resolver/CachingPluginDependencyResolverProvider.kt
fgaraba: It was necessary due to changing null-ness, fixed to an alternative that uses `pluginId`

fgaraba:
> There is no need to address `plugins-verifier-service`. This subproject is essentially archived. I'd revert the changes in that part.

This can't be done unfortunately, as the gradle script compilation itself fails.

novotnyr (approved):

fgaraba:
Kotlin metadata-related changes has been extracted into https://github.com/JetBrains/intellij-plugin-verifier/pull/1591

PR #1561, opened 2026-07-27
Inflatable IDE

novotnyr:
This is a PR based on the `inflatable_ide` branch authored by @fgaraba, rebased on `master`.

### DIFF
--- a/intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/DispatchingIdeManager.kt
+++ b/intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/DispatchingIdeManager.kt
@@ -1,5 +1,6 @@
 package com.jetbrains.plugin.structure.ide
 
+import com.jetbrains.plugin.structure.ide.layout.MissingLayoutFileMode
 import com.jetbrains.plugin.structure.intellij.version.IdeVersion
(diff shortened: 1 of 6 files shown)


intellij-plugin-structure/structure-ide/src/main/java/com/jetbrains/plugin/structure/ide/DispatchingIdeManager.kt
```diff
@@ -13,7 +14,7 @@ class DispatchingIdeManager(configuration: IdeManagerConfiguration = IdeManagerC
   private val standardIdeManager = IdeManagerImpl()
 
   private val productInfoBasedIdeManager = ProductInfoBasedIdeManager(
-    missingLayoutFileMode = configuration.missingLayoutFileMode,
+    missingLayoutFileMode = MissingLayoutFileMode.SKIP_CLASSPATH, // TODO: This should only be changed when inflatable IDE is used
```
novotnyr, line 17: This needs to be set for the traditional value, otherwise there will be breakages.

novotnyr:
We need at least some unit tests that demonstrate
- pure core plugin loading
- single plugin load
- lazyness of the loading process: loading a single plugin, checking the status of the `IDE`, then loading a second one, and checking the status
- lazy loading of two plugins where one of them has transitive deps (that expand the loading subtree).

PR #1567, opened 2026-07-29
Cache ContentModuleScanner results to fix Junie perf issue

fgaraba:
Cache `lib/` scan results in `FileBasedModuleDescriptorResolver` to fix content-module resolution bottleneck                                                                                        
                                                                                                                                                                                                  
Junie (plugin 1072850) declares 64 content modules, none of which use the fast path (lib/modules/<name>.jar) -- they all land in a single JAR under `lib/`. This caused `FileBasedModuleDescriptorResolver.getModuleCreator` to fall through to `LibDirectoryPluginLoader` for every module, which sequentially opens all JARs in lib/ to find the one containing the descriptor. With 243 JARs in Junie's `lib/`, that's 15,552 JAR operations per plugin parse - adding ~2 minutes to structure parsing in the Marketplace's PluginStructureParser.
                                                                                                                                                                                                  
`ContentModuleScanner` already scans all JARs in `lib/` and `lib/modules/` and builds a _moduleId -> artifactPath_ map; it was previously used only for classpath setup. This change reuses it inside `FileBasedModuleDescriptorResolver`: on the first miss of the `lib/modules/<name>.jar` fast path, the resolver calls `ContentModuleScanner` once and caches the result in a map keyed by plugin artifact path. Subsequent modules resolve via O(1) map lookup.

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginManager.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/IdePluginManager.kt
@@ -52,6 +52,7 @@ class IdePluginManager private constructor(
   private val moduleFromDescriptorLoader = pluginLoaderRegistry.get<ModuleFromDescriptorLoader.Context, ModuleFromDescriptorLoader>()
   private val jarModuleLoader = pluginLoaderRegistry.get<JarModuleLoader.Context, JarModuleLoader>()
   private val jarOrDirLoader = pluginLoaderRegistry.get<JarOrDirectoryPluginLoader.Context, JarOrDirectoryPluginLoader>()
+  private val libDirectoryLoader = pluginLoaderRegistry.get<LibDirectoryPluginLoader.Context, LibDirectoryPluginLoader>()
 
   private val contentModuleLoader = ContentModuleLoader(jarOrDirLoader, moduleFromDescriptorLoader)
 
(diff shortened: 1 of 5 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/module/FileBasedModuleDescriptorResolver.kt
```diff
@@ -16,10 +16,17 @@ import com.jetbrains.plugin.structure.intellij.plugin.loaders.JarOrDirectoryPlug
...
 
-internal class FileBasedModuleDescriptorResolver(private val pluginLoader: JarOrDirectoryPluginLoader) :
-  ModuleDescriptorResolver<FileBasedModule>() {
+internal class FileBasedModuleDescriptorResolver(
+  private val pluginLoader: JarOrDirectoryPluginLoader,
+  fileSystemProvider: JarFileSystemProvider
+) : ModuleDescriptorResolver<FileBasedModule>() {
+
+  private val contentModuleScanner = ContentModuleScanner(fileSystemProvider)
+  private val moduleArtifactCache = ConcurrentHashMap<Path, Map<String, Path>>()
```
chrkv: Scope this cache to a single `resolveContentModules` pass rather than the lifetime of `IdePluginManager`. `ContentModuleLoader` is retained by `IdePluginManager`, and the manager can be reused to parse many plugin artifacts, so this map otherwise retains every scanned artifact and its module paths indefinitely.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/plugin/module/FileBasedModuleDescriptorResolver.kt
```diff
@@ -54,6 +64,13 @@ internal class FileBasedModuleDescriptorResolver(private val pluginLoader: JarOr
     ))
   }
 
+  private fun resolveModuleArtifact(pluginArtifactPath: Path, moduleName: String): Path? {
+    val moduleMap = moduleArtifactCache.computeIfAbsent(pluginArtifactPath) { path ->
+      contentModuleScanner.getContentModules(path).modules.associate { it.id to it.artifactPath }
```
chrkv: `associate` silently collapses duplicate module IDs. If the same root-level module descriptor exists in two JARs under `lib/`, we now pick one artifact and bypass `LibDirectoryPluginLoader`, which previously reported `MultiplePluginDescriptors`.

Please preserve all candidates, e.g. as `Map<String, List<Path>>`, and use the direct-JAR fast path only when `singleOrNull()` returns a unique match. For duplicates, fall back to the existing loader or report the ambiguity explicitly.

chrkv (changes requested):

novotnyr (changes requested):
As per @chrkv comments.

novotnyr (approved):

chrkv (approved):
Non-blocking follow-up: the new `getDescriptorProviders()` filters candidates using an index built only from JARs, so ZIPs under `lib/` no longer reach the existing `file.isJar() || file.isZip()` branch for root-level content-module descriptors. ZIPs under `lib/` don't appear to be part of the supported plugin packaging model, so this shouldn't block the PR, but it may be worth tracking separately whether `LibDirectoryPluginLoader` should preserve this historical behavior.

PR #1579, opened 2026-08-11
Use local plugin repositories in online `check-trunk-api` mode

chrkv:
**Description**

`check-trunk-api` accepts `-rjbp` and `-tjbp` repositories with locally built JetBrains plugins, but currently these repositories are ignored unless Plugin Verifier is running in offline mode.

This makes them unavailable during the normal online `check-trunk-api` workflow, where Marketplace is still needed to obtain 3rd-party plugins and dependencies not present locally.

Create the `check-trunk-api` local repositories independently of the global offline mode. Dependency resolution can then prefer locally built JetBrains plugins and fall back to Marketplace when a dependency is missing locally.

Plugins present in the local JetBrains repositories remain excluded from the verification target: `check-trunk-api` verifies whether trunk introduces new problems in 3rd-party plugins rather than verifying the JetBrains plugins themselves.

Added a test covering local dependency resolution with Marketplace fallback in online mode.


novotnyr:
I've removed the reflection in tests with proper usage.
Aside of that, LGTM!

novotnyr (approved):

PR #1583, opened 2026-08-14
WI-85126 suppress internal api usages errors for LaravelIdeaCompatibilityTest

MariyaFilippova:
Laravel Idea is a plugin that lives outside the monorepo but is bundled with PhpStorm. It uses the Internal FUS API, which prevents LaravelIdeaCompatibilityTest from passing.


MariyaFilippova:
@novotnyr can you take a look, please? I can't add a reviewer for some reason

PR #1585, opened 2026-08-19
Validate sinceBuild and untilBuild components more granularly

kesarevs:
JetBrains Marketplace converts the update's since / until values to long using `CompatibilityUtils` to simplify search. This means we need to validate that each meaningful component (the first three components of the IDE version) is within the specified bounds.

See [MP-7678](https://youtrack.jetbrains.com/issue/MP-7678).

### DIFF
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
@@ -107,7 +107,7 @@ class InvalidUntilBuildWithJustBranch(
 class InvalidUntilBuildWithMagicNumber(
   descriptorPath: String,
   untilBuild: String,
-  magicNumber: String
+  magicNumber: Int
 ) : InvalidUntilBuild(
   descriptorPath = descriptorPath,
   untilBuild,
--- a/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
+++ b/intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
@@ -61,8 +61,7 @@ class InvalidSinceBuild(
   sinceBuild: String
 ) : InvalidDescriptorProblem(
   descriptorPath = descriptorPath,
-  detailedMessage = "The <since-build> parameter ($sinceBuild) format is invalid. Ensure it is greater than <130> " +
-                    "and represents the actual build numbers."
+  detailedMessage = "The <since-build> parameter ($sinceBuild) format is invalid. Ensure it represents the actual build numbers."
 ) {
   override val level
     get() = Level.ERROR
@@ -163,6 +162,25 @@ class ErroneousSinceBuild(
 ) {
   override val hint = ProblemSolutionHint(
     example = "since-build=\"182.4132.789\"",
+    documentationUrl = "https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html"
+  )
+
+  override val level: Level
+    get() = Level.ERROR
+}
+
+class IdeBuildComponentsOutOfRange(
+  ideVersion: IdeVersion,
+  failedComponent: Int,
+  range: IntRange,
+  attributeName: String,
+  descriptorPath: String
+) : InvalidDescriptorProblem(
+  descriptorPath = descriptorPath,
+  detailedMessage = "The `$failedComponent` component of the <$attributeName> parameter ($ideVersion) is out of range [${range.first}; ${range.last}]."
+) {
+  override val hint = ProblemSolutionHint(
+    example = "$attributeName=\"182.4132.789\"",
     documentationUrl = "https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html"
   )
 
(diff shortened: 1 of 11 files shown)


intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
```diff
@@ -61,8 +61,7 @@ class InvalidSinceBuild(
   sinceBuild: String
 ) : InvalidDescriptorProblem(
   descriptorPath = descriptorPath,
-  detailedMessage = "The <since-build> parameter ($sinceBuild) format is invalid. Ensure it is greater than <130> " +
-                    "and represents the actual build numbers."
+  detailedMessage = "The <since-build> parameter ($sinceBuild) format is invalid. Ensure it represents the actual build numbers."
```
novotnyr, line 64: The '130' is a magic number, but is there a reason to remove it?

novotnyr (approved):
While reviewing, I have

- restored the magic number (999 and alike) compatibility to original behaviour
- added 0-specific tests

Looking good to me!

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/problems/InvalidDescriptorErrors.kt
kesarevs: Yes, it's now checked in

```
verifyIdeBuildComponentsRanges(
  ideVersion = sinceBuildParsed,
  baselineLowerBound = SINCE_BASELINE_LOWER_BOUND,
  attributeName = IdeaVersionBean.SINCE_BUILD_ATTRIBUTE_NAME,
  descriptorPath = descriptorPath
)
```

where SINCE_BASELINE_LOWER_BOUND = 130.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/verifiers/PluginSinceUntilRangeVerifier.kt
```diff
@@ -142,7 +155,7 @@ class PluginSinceUntilRangeVerifier {
     val baseline = untilBuild.baselineVersion
     if (baseline >= FIRST_YEARLY_BASED_RELEASE_NUMBER_YEAR) {
       registerProblem(InvalidUntilBuild(descriptorPath, untilBuildValue, untilBuild))
-    } else if (baseline == MAGIC_BASELINE_NUMBER) {
+    } else if (baseline >= MAGIC_BASELINE_NUMBER) {
```
kesarevs: @novotnyr Is there a reason why you need `>=` here? We already checked that baseline should be `< 1000` in `verifyIdeBuildComponentsRanges`. And 1000, 1001, ... -- are not `magic numbers`, they are out of bounds.

intellij-plugin-structure/structure-intellij/src/main/java/com/jetbrains/plugin/structure/intellij/verifiers/PluginSinceUntilRangeVerifier.kt
```diff
@@ -107,6 +107,11 @@ class PluginSinceUntilRangeVerifier {
       return
     }
 
+    untilBuild.magicNumberSingleComponent()?.let { magicNumber ->
+      registerProblem(InvalidUntilBuildWithMagicNumber(descriptorPath, untilBuild, magicNumber))
+      return
+    }
```
kesarevs: @novotnyr Do we really need this? Single component build is valid as `IdeVersion`. And magic numbers will be checked in `verifyUntilBuildBaselineMagicNumbers`

kesarevs:
> While reviewing, I have
> 
> * restored the magic number (999 and alike) compatibility to original behaviour
> * added 0-specific tests
> 
> Looking good to me!

If I'm not mistaken only `0` in all components case wasn't covered by the branch. I think it might be easier and more clear to add the check into `verifyUntilBuildBaselineMagicNumbers` (and rename it).

As for the `999` magic number and others it was covered by the initial PR. Please correct me if I'm wrong.

novotnyr:
> As for the `999` magic number and others it was covered by the initial PR. Please correct me if I'm wrong.

Originally, any number >= 999 was considered to be magic. Then, after the changes, only sharp 999 is magic. I have restored the original semantics there.

novotnyr:
> If I'm not mistaken only `0` in all components case wasn't covered by the branch. I think it might be easier and more clear to add the check into `verifyUntilBuildBaselineMagicNumbers` (and rename it).

We have multiple tests, and this seems to be all just corner-cases, so we are OK, IMHO.

kesarevs:
> > As for the `999` magic number and others it was covered by the initial PR. Please correct me if I'm wrong.
> 
> Originally, any number >= 999 was considered to be magic. Then, after the changes, only sharp 999 is magic. I have restored the original semantics there.

Well, that is the idea of the refactoring, to drop everything we don't need anymore, right? Since from now on all builds with baseline `> 999` will be considered invalid with `IdeBuildComponentsOutOfRange` exception, I think it's totally enough.

Anyway, `verifyUntilBuildBaselineMagicNumbers` is still in use and all real magic numbers are still covered. At the same time, I don't think it's a good practice to show user an error about some `magic values` if it's just out of bounds.

kesarevs:
Sorry for reverting part of your changes.

The >= 999 behavior was intentionally changed, so I don't think restoring it is the right direction. I also simplified the 0 case a bit while keeping the coverage and tests.

For behavioral changes like this, I'd prefer to discuss them first rather than commit them directly to the branch.

novotnyr:
Sure, let's discuss it in the future PRs! Now it's clear!

Thank you for the contribution, LGTM.

novotnyr (approved):

